PR #156, opened 2014-12-25
Changes for the next Mayavi release which uses VTK's new pipeline.

dmsurti:
1. Bump up to version 5.0.0.
2. Document all enhancements and fixes.
3. Add self name as contributor.

### DIFF
--- a/docs/CHANGES.txt
+++ b/docs/CHANGES.txt
@@ -1,4 +1,4 @@
-Mayavi 5.0.0
+Mayavi 4.4.0
 ============
 
 Enhancements
--- a/docs/CHANGES.txt
+++ b/docs/CHANGES.txt
@@ -4,23 +4,23 @@ Mayavi 5.0.0
 Enhancements
 ------------
 
-22 Dec 2014 (PR)
+22 Dec 2014 (DS)
     - Add function to set data at input port, add stanford (bunny,
       dragon, lucy) examples, and use new volume mapper for new pipeline.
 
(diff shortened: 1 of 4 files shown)


docs/CHANGES.txt
```diff
@@ -1,3 +1,44 @@
+Mayavi 5.0.0
+============
+
+Enhancements
+------------
+
+22 Dec 2014 (PR)
```
prabhuramachandran: Thanks but, these changes are by DS and not PR. :)

mayavi/__init__.py
```diff
@@ -5,7 +5,7 @@
     Part of the Mayavi project of the Enthought Tool Suite.
 """
 
-__version__ = '4.3.2'
+__version__ = '5.0.0'
```
prabhuramachandran: Why not 4.4?

dmsurti: As we upgrade from VTK 5.10.1 to VTK 6, especially using the new pipeline architecture, I thought this qualified for major version upgrade.

GaelVaroquaux: Hey Deepak,

Thanks for doing all this! It's really great to know that Mayavi
development is alive and kicking! This software is really widely used.

docs/CHANGES.txt
```diff
@@ -1,3 +1,44 @@
+Mayavi 5.0.0
```
prabhuramachandran: This needs to be changed.

PR #172, opened 2015-04-22
Fix OS X trackpad scrolling bug by throttling wheel events.

dmsurti:
This commit works around a qt bug, affecting scrolling on trackpad on mac.
(https://bugreports.qt-project.org/browse/QTBUG-22269)

QVTKRenderWindowInteractor fails to keep up with the rate at which QWheelEvents come in.
We accumulate them and send them once every 25ms.

### DIFF
--- a/tvtk/pyface/ui/qt4/QVTKRenderWindowInteractor.py
+++ b/tvtk/pyface/ui/qt4/QVTKRenderWindowInteractor.py
@@ -226,6 +226,14 @@ class QVTKRenderWindowInteractor(QtGui.QWidget):
         self._Timer = QtCore.QTimer(self)
         self.connect(self._Timer, QtCore.SIGNAL('timeout()'), self.TimerEvent)
 
+        # add wheel timer to fix scrolling issue with trackpad
+        self.wheel_timer = QtCore.QTimer()
+        self.wheel_timer.setSingleShot(True)
+        self.wheel_timer.setInterval(25)
+        self.wheel_timer.timeout.connect(self._emit_wheel_event)
+        self.wheel_accumulator = 0
+        self._saved_wheel_event_info = ()
+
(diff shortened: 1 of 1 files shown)


prabhuramachandran:
Great!  Thanks!

tvtk/pyface/ui/qt4/QVTKRenderWindowInteractor.py
```diff
@@ -424,11 +432,37 @@ def keyReleaseEvent(self, ev):
...
+        """ Reimplemented to work around scrolling bug in Mac.
+
+        Work around https://bugreports.qt-project.org/browse/QTBUG-22269.
+        Accumulate wheel events that are within a period of 25ms into a single
+        event.  Changes in buttons or modifiers, while a scroll is going on,
+        are not handled, since they seem to be too much of a corner case to be
+        worth handling.
+        """
+        self.wheel_accumulator += ev.delta()
+        self._saved_wheel_event_info = (
```
pankajp: I think you should accumulate the events only when they are similar. I.e. check that atleast the buttons, modifiers and orientation of the new event match the properties of events accumulated upto now, otherwise diapatch the accumulated one and start accumulating the new event from delta=0.

dmsurti: Is this detailed checking really required as the wheel events are generated only for mouse wheels and trackpad scroll gestures? So throttling for both should be just fine.

pankajp: Well Ctrl+Scroll or etc have different in many apps. If VTK does not provide a way to handle them then i guess it is not needed, but in the general case it is.
For example, in browsers, Cmd+scroll usually zooms, so there would be upto 25ms error in handling switching. Maybe it does not matter

Issue #190, opened 2015-06-08
support mayavi build using vtk 6.2


itziakos:
When bulding mayavi on top of vtk 6.2 we get exceptions from the vtk_parser.

```python
$ python setup.py build
running build
----------------------------------------------------------------------
Building TVTK classes...
Traceback (most recent call last):
  File "setup.py", line 471, in <module>

  File "lib/python2.7/site-packages/numpy/distutils/core.py", line 169, in setup
    return old_setup(**new_attr)
  File "lib/python2.7/distutils/core.py", line 152, in setup
    dist.run_commands()
  File "/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "setup.py", line 303, in run

  File "setup.py", line 290, in build_tvtk_classes_zip

  File "tvtk/setup.py", line 62, in gen_tvtk_classes_zip
    # numpy.distutils does the Wrong(TM) thing.
  File "tvtk/code_gen.py", line 101, in generate_code
  File "tvtk/code_gen.py", line 191, in _write_wrapper_class
  File "tvtk/wrapper_gen.py", line 80, in generate_code
  File "tvtk/wrapper_gen.py", line 180, in _gen_methods
  File "tvtk/wrapper_gen.py", line 745, in _gen_other_methods
  File "tvtk/wrapper_gen.py", line 1046, in _write_tvtk_method
  File "tvtk/vtk_parser.py", line 309, in get_method_signature
AttributeError: 'NoneType' object has no attribute 'split'
```
I saw that it failed for when the method argument to get_method_signature was a type object.

jenshnielsen:
I guess this could be the same issue as #159 ?

itziakos:
True, the issues look similar and it might be they are cause by the same bug. @dmsurti we should investigate both.

dmsurti:
@jenshnielsen @itziakos The issues are caused by the same underlying bug for VTK 6.2. This is due to changes in VTK api which need to be handled on the VTK parser that generates the TVTK bindings.

dmsurti:
Fixed in #191

Issue #159, opened 2015-01-30
Mayavi installation fails: 'NoneType' object has no attribute 'find'


hohlraum:
Hi,

I have working copies of the dependencies (vtk, traits, etc...), now I'm trying to install Mayavi. The process fails when using both pip or setup.py. The tail of the traceback is included below. I can include the whole thing (it's very long) if it would help. A debug postmortem shows that it fails on the method ```EventIds```

I'm running this under Mac OSX 10.9.5, with Python 2.7.8

```      
...
File "tvtk/wrapper_gen.py", line 734, in _gen_other_methods
  self._write_tvtk_method(out, vtk_meth)
File "tvtk/wrapper_gen.py", line 1035, in _write_tvtk_method
  sig = self.parser.get_method_signature(vtk_meth)
File "tvtk/vtk_parser.py", line 304, in get_method_signature
  doc = doc[:doc.find('\n\n')]
AttributeError: 'NoneType' object has no attribute 'find'
```

ariell:
I am having the same issues. Any resolution on this?

myfavecolorisgreen:
I'm having the same problem too. I think mayavi installed okay the first time I tried pip install mayavi, but I hadn't correctly installed vtk at that point, so import mayavi did not work. After getting vtk installed correctly, I ran pip uninstall mayavi, thinking that I should try to reinstall it. Now, I'm not able to install mayavi, and get the same error as above. 

I have tried pip uninstall-ing all of mayavi's dependencies and then running pip install mayavi again, but this didn't work either.

ElliotKaplan:
I had the same problem (Python 2.7.8, Ubuntu 14.04). The workaround that I settled on was to catch the error in  _write_tvtk_method and ignore it

        if not sig:
            try:
                sig = self.parser.get_method_signature(vtk_meth)
            except:
                print 'err159\t', vtk_meth
                return

The build trace can be seen here http://pastebin.ca/2985442
This resulted in a mostly usable version of Mayavi2---vector and scalar cut planes are not working properly, which may be a function of something else---though I'm not sure exactly what's being swept under a rug here. The print statement gets triggered for a bunch of <type 'SomeType'> objects.

ckalas:
@ElliotKaplan thanks, this allowed me to finally build it too.

jjstickel:
A patch for the workaround described by ElliotKaplan can be found here:
https://trac.macports.org/attachment/ticket/47762/mayavi_wrapper.patch

[...]

PR #191, opened 2015-06-09
Ignore false positives for built in methods/functions when parsing

dmsurti:
This implements support for #190 (and hence #159).

* When generating the TVTK bindings for VTK python wrappings, for VTK version 6.2, due to the API changes, a few additions (example: EventIds in vtkCommand and DesiredOutputPrecision in vtkAlgorithm) are falsely deteced as methods to be parsed. These are ignored when generating the tvtk methods.
* The method signature implementation now checks for VTK version 6.2 and ignores those methods which are not built in functions or methods. The method that writes out the TVTK method then ignores such signatures which are `None`.
* Tested on OS X 10.10: `python setup.py build` with VTK 6.2 egg now runs without errors.
* Tested for backwards compatibility with VTK 6.1 and VTK 5.10.

### DIFF
--- a/tvtk/wrapper_gen.py
+++ b/tvtk/wrapper_gen.py
@@ -1046,7 +1046,8 @@ class WrapperGenerator:
         if sig is None:
             sig = self.parser.get_method_signature(vtk_meth)
         
-        # VTK 6.2: When the method signature is None, ignore the method
+        # VTK 6.2: There exists no method signature for false built in
+        # functions/methods
         if sig is None:
             return
 
--- a/tvtk/wrapper_gen.py
+++ b/tvtk/wrapper_gen.py
@@ -16,7 +16,7 @@ import copy
 
 # Local imports (these are relative imports because the package is not
 # installed when these modules are imported).
-from common import get_tvtk_name, camel2enthought
+from common import get_tvtk_name, camel2enthought, is_version_62
 import vtk_parser
 import indenter
 import special_gen
@@ -744,6 +744,7 @@ class WrapperGenerator:
             vtk_meth = getattr(klass, m)
             self._write_tvtk_method(out, vtk_meth)
 
+
     #################################################################
     # Private utility methods.
     #################################################################
@@ -1044,6 +1045,9 @@ class WrapperGenerator:
         """
         if not sig:
             sig = self.parser.get_method_signature(vtk_meth)
+            # VTK 6.2 false built in funcs/methods are ignored
+            if sig is None:
+                return
 
         # Figure out if we really need to wrap the return and deref
         # the args.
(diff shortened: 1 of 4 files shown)


tvtk/wrapper_gen.py
```diff
@@ -1044,6 +1045,9 @@ def _write_tvtk_method(self, out, vtk_meth, sig=None):
         """
         if not sig:
```
itziakos: I think that this code block should be:

```python
if sig is None:
    sig = self.parser.get_method_signature(vtk_meth)
if sig is None:
    # When the signature is None then the method should be ignored
    return
```

prabhuramachandran: I agree that this is easier to read although I do not believe there should be a problem currently as if the user explicitly passed an empty signature string, then clearly that is what they want.  In any case, I think this is safe to change.

tvtk/wrapper_gen.py
```diff
@@ -1042,8 +1043,12 @@ def _write_tvtk_method(self, out, vtk_meth, sig=None):
           If None, this is computed.  If not, the passed signature
           information is used.
         """
-        if not sig:
+        if sig is None:
             sig = self.parser.get_method_signature(vtk_meth)
+        
+        # VTK 6.2: When the method signature is None, ignore the method
```
itziakos: I think that the comment is a little misleading, since the check does not know why the get_metho_signature returned None. We only know that when signature is None we ignore the method from wrapping.

itziakos:
Can we add a small test to check that we will not attempt to wrapp a 'built-in method' in 6.2. The tests can be skipped when vtk version is not 6.2.

itziakos:
Please update the changelog with the necessary info.

dmsurti:
This test https://github.com/enthought/mayavi/blob/master/tvtk/tests/test_vtk_parser.py#L274 fails without the fixes in this PR. So testing is already handled. In fact, I had to rework my initial implementation and move code to `get_method_signature` to handle the failing test.

dmsurti:
At the time of release, I have followed the current Mayavi process of releasing 2 PR's: one for change logs, bumping up version and the other for updated cos. Not sure about updating change log for each / this PR.

dmsurti:
Pending my feedback comments, I think good to start making a release.

prabhuramachandran:
LGTM, merging.

catcatcz:
Hi, I was trying to install mayavi on linux Centos. To do this, I first installed VTK 6.3.0. Then I cd into mayavi-4.4.3 that is unzipped from mayavi-4.4.3.zip, run sudo python setup build. It reports the error: " numpy.disutils does the wrong thing" and "Attribuite error:'Nonetype' object has no attribute 'find'. " 

I do not know whether it has the same original bug as this one in vtk 6.2, although they look similar. I am asking for help from you guys... Thank you.

jenshnielsen:
@catcatcz There is a pull request to add support for VTK 6.3 in #247 With that one I was able to install Mayavi on top of 6.3 but without I had a similar error as you describe.

PR #196, opened 2015-06-12
bump version to 4.4.3.dev0

itziakos:


itziakos:
@dmsurti is it ok to merge?

PR #199, opened 2015-06-18
Add appveyor CI runs for windows tests

itziakos:
This PR provides basic CI testing on windows using appveyor:

In more details:
- appveyor is setup to run on commited changes and PRs
- dependencies like numpy are cached 
- vtk 6.2 with python is downloaded from http://www.vtk.org/download/, The python package is installed (hacked) into the system python using the `PYTHONPATH`
- no integration tests are run because the build freezes
- fixed issues with instantiating:
  - vtkHardwareSelection and vtkOpenGLHardwareSelector on 6.2.
  - vtkAxesTransformRepresentation on 5.8.


itziakos:
there is a strange failure 

```python
ERROR: Test if all the TVTK classes are instantiable.
  File "C:\projects\mayavi\tvtk\tests\test_tvtk.py", line 621, in test_all_instantiable
    "Errors occured during this test, see printed messages."
TraitError: Errors occured during this test, see printed messages.
-------------------- >> begin captured stdout << ---------------------
class: HardwareSelector The 'prop_color_value' trait of a HardwareSelector instance must be a tuple of the form: (0.0 <= a floating point number <= 1.0, 0.0 <= a floating point number <= 1.0, 0.0 <= a floating point number <= 1.0), but a value of (0.0, 1.875, 0.0) <type 'tuple'> was specified.
 
--------------------- >> end captured stdout << ----------------------
 
Ran 84 tests in 28.431s
 
FAILED (errors=1)
Command exited with code 1
```

itziakos:
it looks that the problems with `HardwareSelector` is that the `PropColorValue` is not initialized as a result the value going to be based on the allocated memory and thus will not always conform to the 0 - 1.0 range.

itziakos:
@dmsurti, I have fixed the issues with instantiating tvtk classes and cleanup a little the `tvtk_tests.py`. Please have a look. One item that I am not sure about is the support for python 2.6.

dmsurti:
@itziakos LGTM. Is it ready for merge? 

Except: A minor observation / query: I think we must eventually use VTK eggs on our CI infrastructure.

itziakos:
@dmsurti, I agree that using the vtk provided python distribution is not ideal. The best solution would be to use wheels for vtk on pypi (e.g. pyside has such wheels).

PR #209, opened 2015-06-30
Use travis ci containers

itziakos:
This PR converted the travis-ci tests to use the new travis-ci containers so that the builds start faster.

In more details:
- Tests are run on both qt and wx
- Coverage is now uploaded onto codecov.io (https://codecov.io/github/enthought/mayavi)
- caching is used to speed up builds

**note**: The integration tests hang with Qt so they are skipped

### DIFF
--- a/travis-install-wx.sh
+++ b/travis-install-wx.sh
@@ -1,7 +1,6 @@
 #!/bin/bash
 set -e
-rm -Rf wxPython-src-2.8.12.1
-wget http://sourceforge.net/projects/wxpython/files/wxPython/2.8.12.1/wxPython-src-2.8.12.1.tar.bz2/download
+wget http://sourceforge.net/projects/wxpython/files/wxPython/2.8.12.1/wxPython-src-2.8.12.1.tar.bz2/download && md5sum -c checksums
 tar -xf download
 cd wxPython-src-2.8.12.1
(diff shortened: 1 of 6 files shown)


itziakos:
@dmsurti this PR is ready for review

travis-install-wx.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -e
+rm -Rf wxPython-src-2.8.12.1
+wget http://sourceforge.net/projects/wxpython/files/wxPython/2.8.12.1/wxPython-src-2.8.12.1.tar.bz2/download
```
dmsurti: Is it a good idea to download from `sourceforge` with reports of sourceforge bundling malware in the downloads?

dmsurti:
Other than the comment about `source forge` downloads, LGTM.

travis-install-wx.sh
itziakos: I think that the sourceforge problem does not apply here.

- sourceforge is the official repo for wxPython unless this changes we have to download from there.  
- we do not download a bundled binary installer, but a zipped source file. 
- the package is installed in a container and then discarded.

jenshnielsen: In any case wouldn't it be a good idea to do a checksum verification?

itziakos: That is a good point

itziakos: I have added a checksum check. However please note that since I could not find official checksums for the wxPython source zip I used the source zip that I downloaded locally. This will only protect against future changes of the file.

PR #212, opened 2015-07-08
Fix memory leaks bug - v1

dmsurti:
This is first of a set of fixes to fix memory leaks. The fixes will be rolled out gradually in different milestones.

This removes the reference cycles between Scene, Light Manager, Picker and VTKRenderWindowInteractor. It also invokes closed event handlers when the scene is closed.


prabhuramachandran:
Awesome, can you please add a simple test case for this?  Thanks!

tvtk/pyface/ui/qt4/scene.py
```diff
@@ -338,7 +339,7 @@ def __init__(self, parent=None, **traits):
         super(Scene, self).__init__(parent, **traits)
 
         # Setup the default picker.
-        self.picker = picker.Picker(self)
+        self.picker = picker.Picker(weakref.proxy(self))
```
itziakos: Is there a reason why we do not change how `Picker` keeps a reference to the scene? (See line https://github.com/dmsurti/mayavi/blob/fix-memory-leaks-bug/tvtk/pyface/picker.py#L242). I also wonder if `renwin` of Picker could be a WeakRef trait.

tvtk/pyface/ui/qt4/scene.py
```diff
@@ -394,14 +395,25 @@ def save_to_clipboard(self):
...
     # Non-public interface.
     ###########################################################################
     def _create_control(self, parent):
         """ Create the toolkit-specific control that represents the widget. """
 
         # Create the VTK widget.
-        self._vtk_control = window = _VTKRenderWindowInteractor(self, parent,
-                                                                 stereo=self.stereo)
+        proxy_scene = weakref.proxy(self)
+        self._vtk_control = window = _VTKRenderWindowInteractor(proxy_scene,
```
itziakos: Similar comment with https://github.com/enthought/mayavi/pull/212/files#r34133394

tvtk/pyface/tvtk_scene.py
```diff
@@ -808,7 +817,7 @@ def _create_control(self, parent):
 
         self._interactor.initialize()
         self._interactor.render()
-        self.light_manager = light_manager.LightManager(self)
+        self.light_manager = light_manager.LightManager(weakref.proxy(self))
```
itziakos: similar comment https://github.com/enthought/mayavi/pull/212/files#r34133394. Could `LightManager` handle the weak proxy ref internarly?

tvtk/pyface/tvtk_scene.py
```diff
@@ -228,6 +229,14 @@ def __setstate__(self, str_state):
         state_pickler.set_state(self, state_pickler.loads_state(str_state))
 
     ###########################################################################
+    # 'event' interface.
+    ###########################################################################
+    def _closed_fired(self):
+        del self.light_manager
+        del self._interactor
+        del self
```
itziakos: I think for traits it is better to do.

```python
self.light_manager = None
self._interactor = None
```

Also there should not be a need to `del self` since `self` will be garbadge collected as soon as the method is called,

tvtk/pyface/ui/qt4/scene.py
```diff
@@ -394,14 +395,25 @@ def save_to_clipboard(self):
         os.unlink(name)
 
     ###########################################################################
+    # 'event' interface.
+    ###########################################################################
+    def _closed_fired(self):
+        del self.picker
```
itziakos: see similar comment in https://github.com/enthought/mayavi/pull/212/files#r34133840

itziakos:
We probably need to check the wx scene for similar problems

dmsurti:
@prabhuramachandran Added a simple test with multiple sample runs. With full memory leaks, the leaks should tend to zero :-). https://github.com/enthought/mayavi/commit/e4f245fe93a744ea24feecf5b89fa50e1dd4b3d5#diff-034e39dcece480fc50d0d9124fa0e684R54

tvtk/pyface/ui/qt4/scene.py
dmsurti: I used this idiom of passing a `weakref proxy` to make it explicit when the relationship is set.

itziakos: I would think that making using a weak ref inside an __init__ method is less explicit than using a WeakRef trait in Picker. It will also mean that we avoid cycles between the scene and the picker in general. For example through out the life of the Scene object the picker could be replaced 

```python
my_picker = Picker(scene)
scene.picker = my_picker
```

tvtk/tests/test_tvtk.py
```diff
@@ -200,7 +200,7 @@ def test_object_cache(self):
         # The test sometimes fails as VTK seems to generate objects with the
         # same memory address and hash, we try to force it to allocate more
         # objects so as to not end up reusing the same address and hash.
-        junk = [vtk.vtkConeSource() for i in range(5)]
+        junk = [tvtk.ConeSource() for i in range(50)]
```
itziakos: what is the purpose for increasing this range?

dmsurti: One of those unreliable test which started failing. Changing to `tvtk` object and allocating more objects (hence increased the range) results in different address and hash.

tvtk/pyface/ui/qt4/scene.py
dmsurti: I have added an issue https://github.com/enthought/mayavi/issues/214 for this as I do not know why `renwin`, and other `pickers (point picker, cell picker etc)` are not traits. @prabhuramachandran ?

itziakos: I still believe that the weak reference binding should happen at the `Picker`.

integrationtests/mayavi/test_memory_usage.py
```diff
@@ -0,0 +1,215 @@
...
+        cp.actor = cp.actor.__class__()
+
+        s.render()
+
+        # Now check.
+        self.check()
+
+
+        ############################################################
+        # Test if saving a visualization and restoring it works.
```
itziakos: Is this part of the code necessary?

itziakos:
It is not easy to understand what the test is actually testing. Do we really need to perform all these operations (e.g. load/save, manipulate contrours) to flesh out the memory leak?  I would also expect that only the memory assertions are necessary.

integrationtests/mayavi/test_memory_usage.py
dmsurti: Yes, I thought it a good idea to do the checks for memory usage with saving/restoring a visualization.

dmsurti:
The idea for this test case it to check for memory usage with a typical workflow. Hence I chose an integration test to add memory checks.

itziakos:
Without a more targeted test it is not easy to see what has been fixed.

itziakos:
Actually since that this memory leak is due to a reference cycle. We should be able to write a test that can verify that the reference cycle is not there anymore using the standard `gc` module. (see examples https://github.com/enthought/traits/blob/d79baa40f43c0482080f2b345b8c201881823599/traits/tests/test_dynamic_notifiers.py#L225 and https://github.com/enthought/traits/blob/d79baa40f43c0482080f2b345b8c201881823599/traits/tests/test_dynamic_notifiers.py#L249)

tvtk/tests/test_scene.py
```diff
@@ -0,0 +1,84 @@
...
+        del a
+        self.assertEqual(self.object_collected, [True])
+        self.assertEqual(r(), None)
+
+    def test_tvtk_scene_garbage_collected(self):
+        """ Test the garbage collection of TVTScene object."""
+        scene = TVTKScene()
+        scene_weakref = weakref.ref(scene, self.object_collected_callback)
+        del scene
+        ################################################################
```
itziakos: I would remove the `gc.collect` and designate these test methods as expected failure. Currently the tests will pass independent if we break the reference cycles or not.

itziakos: Actually the `gc.collect` should happen at the beginning of the test method to make the test more robust. Let me work with them.

tvtk/tests/test_tvtk.py
itziakos: Reading the comment before the specific line of code looks that there is a bug somewhere between the tvk and tvtk layer. Is there a related bug issue for this problem?

appveyor.yml
```diff
@@ -14,6 +14,7 @@ cache:
 
 init:
   - ps: $Env:path = $Env:python + ";" + $Env:python + "\scripts;" + "C:\Program Files\Microsoft SDKs\Windows\v7.0\Bin;" + "C:\vtk\bin;" + $Env:path
+  - ps: $env:ETS_TOOLKIT="qt4"
```
itziakos: please move this into the global section of the environment variables.

itziakos: Actually, this line should not be needed. It looks that the new `test_scene` is doing things wrong.

tvtk/tests/test_scene.py
```diff
@@ -0,0 +1,84 @@
...
+        # Without forcing garbage collection, TVTKScene is not collected
+        ################################################################
+        gc.collect()
+        self.assertEqual(self.object_collected, [True])
+        self.assertEqual(scene_weakref(), None)
+        self.assertEqual(gc.collect(), 0)
+    
+    @unittest.skipIf(os.environ['ETS_TOOLKIT'] == 'wx',
```
itziakos: This should check the ETSConfig class from traits something like
```
ETSConfig.toolkit == 'wx'
```

tvtk/tests/test_scene.py
```diff
@@ -0,0 +1,84 @@
...
+        del scene
+        ################################################################
+        # Without forcing garbage collection, Scene is not collected
+        ################################################################
+        gc.collect()
+        self.assertEqual(self.object_collected, [True])
+        self.assertEqual(scene_weakref(), None)
+        self.assertEqual(gc.collect(), 0)
+
+    @unittest.skip("Skip wx test as it causes segmentation fault")
```
itziakos: Do we know why this happens, Is there a bug report?

integrationtests/mayavi/common.py
```diff
@@ -316,6 +316,9 @@ class TestCase(Mayavi):
     # `self.compare_image` was called in the test..
     offscreen = Bool(False)
 
+    # Profile for memory usage.
+    profile = Bool(False)
```
itziakos: The changes in the common.py do not perform memory profiling, but check for memory usage. Using the word `profile` is misleading.

integrationtests/mayavi/common.py
```diff
@@ -383,12 +386,18 @@ def run_standalone(self):
...
+
     def run(self):
         """This starts everything up and runs the test.  Call main to
         run the test."""
         # Calls the users test code.
         try:
-            self.do()
+            if self.profile:
+                memory_assistant = MemoryAssistant()
+                memory_assistant.assertReturnsMemory(self.do_profile,
```
itziakos: As a design idea there is no reason to have a separate method for the `do_profile` if the user want to run the integration test and check for reference cycles then just run the normal test `do` using the memory_assistant.

integrationtests/mayavi/test_contour.py
```diff
@@ -177,6 +177,68 @@ def do(self):
 
         # If we have come this far, we are golden!
 
+    def do_profile(self):
```
itziakos: This code does not check the specifics changes that are put forward by the this PR.

tvtk/tests/test_tvtk.py
dmsurti: This is related to an issue with deleting vtk objects, which has been fixed. See: http://review.source.kitware.com/#/c/15095/. This test is flaky on VTK 5.x, not on 6.x. Not necessary to add an issue ticket for this.

Other reference: https://github.com/enthought/mayavi/pull/138.

itziakos:
@dmsurti, I had a look in the new changes. In general a memory usage check framework for the integration tests is a good idea which we can extent in the future but we still need improve this specific PR to check and verify that the provided changes remove some of the reference cycles in an automatic  way. 

Let me spend some to make the necessary changes

tvtk/tests/test_scene.py
dmsurti: Added issue: https://github.com/enthought/mayavi/issues/216. I don't know why this happens.

integrationtests/mayavi/common.py
dmsurti: The reason to have a separate method `do_profile` is that the the method `do` also does a series of checks/assertions. We do not want to do that while checking for memory usage/leaks. In the earlier version of `test_memory_usage`, we found that reusing `do` for memory usage etc resulted in less readable code due to the assertions, saving/loading vis etc.

integrationtests/mayavi/test_contour.py
dmsurti: The intent is to add a framework for memory checks by reusing integration suite. So in that context, it does not check the specifics.

integrationtests/mayavi/common.py
dmsurti: What is a good name you suggest? I can think of `check_mem_usage`. 

Any references to explain why `profile` is misleading? I am only a novice when tackling memory issues :-).

integrationtests/mayavi/common.py
itziakos: I am discussing the design extension of the integration tests framework. It looks weird that one has to replicate almost the same code once in do and once do_profile. A better api would be that we always perform `self.do()` but we use the memory assistant when the usecase is designated as a memory leak check. 

In short we should keep the `test_contour.py` and we add another called `test_contour_memory.py` (or in the same file but separate test cases).

dmsurti: As I mentioned earlier, `do` does a lot more stuff.

As per your earlier comment (quoting only for reference):
--------
It is not easy to understand what the test is actually testing. Do we really need to perform all these operations (e.g. load/save, manipulate contrours) to flesh out this specific memory leak? I would also expect that only the memory assertions are necessary.
--------

Hence a separate do_profile method. Separating is also not a great idea as then we have to classify between integration/memory tests. 

Alternative approach I can think of is: to refactor `do` into those separate methods and ignore those which are not relevant to memory checks.

integrationtests/mayavi/common.py
itziakos: Two main reasons:
(i) just `profile` can mean memory or time profiling
(ii) profiling in general means collecting detailed information during the code run (e.g see https://pypi.python.org/pypi/memory_profiler, python profiler). In our case we are not creating a memory profile of the run over time (or lines or function calls) but we are reporting and checking the memory usage in a much more crude way.

dmsurti:
Implemented in #218. Closed.

PR #215, opened 2015-07-13
Added setting of QVariant for example to work with PyQt4

mwilmer:
Hi Guys,

The qt embedding example with PyQt didn't work for me. Not on my mac, neither on my linux machines. I think it will also hold for others. Apparently, the QVariant also needs to be set before loading the PyQt modules. Couldn't find any other way to let you know, so I added this to the example here so it's easy to change.

kinds regards,

Mike Wilmer

### DIFF
--- a/docs/source/mayavi/auto/qt_embedding.py
+++ b/docs/source/mayavi/auto/qt_embedding.py
@@ -21,6 +21,7 @@ from pyface.qt import QtGui, QtCore
 # the following lines are executed before the import of PyQT:
 #   import sip
 #   sip.setapi('QString', 2)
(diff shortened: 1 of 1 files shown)


prabhuramachandran:
Hi Mike,

Thanks for this, we are not sure about the exact use case.  On Travis we are testing Mayavi on PyQt4 and there have been no problems, can you give us a little more context on this error?  Which version of PyQt is this?  Thanks!

Prabhu

prabhuramachandran:
@mwilmer -- am not sure this is really needed and as I see no responses I am closing this for now.  Please reopen if you need to.  Thanks!

PR #217, opened 2015-07-15
Fix #213: the VTKFileReader when using standalone output not updated

### LINKED ISSUE #213: VTKFileReader source output not set properly
It looks that the VTKFileReader outputs attribute is not set to the right value. Here is an example using mayavi 4.4.2 and vtk 6.2.0 and the vtk.vtk file from the VTKData repo (http://www.vtk.org/Wiki/VTK_Datasets)

```python
from mayavi.sources.vtk_file_reader import VTKFileReader

source = VTKFileReader()
source.initialize('vtk.vtk')

print source.outputs[0]  # this dataset does not contain the 12 points and 5 lines stored in vtk.vtk.
print source.reader.output
```

Please note that when loading the file in mayavi2 and adding modules for visualization the pipeline gets the right info.

dmsurti:
Fixes #213.

When using stand alone, the assign attribute filter is not updated which results in file reader output not updated.

1. A test case has been added which test various datasets.
2. This file reader works fine in a pipeline and is tested in integration
tests.

Please note that if a `*.vtk` file exists without point or cell data, then the assign attribute will throw a harmless warning as such which can be safely ignored:

`vtkAssignAttribute (0x120b38760): Data must be point or cell for vtkDataSet`

This warning results from: https://gitlab.kitware.com/vtk/vtk/blob/master/Filters/Core/vtkAssignAttribute.cxx#L289


dmsurti:
@itziakos @prabhuramachandran The test `test_vtk_file_reader.test_unstructured_grid_file` that I added passes on AppVeyor. But it fails on Travis. The sample file `UGridVTK.ext` has been added, but on Travis it fails with:
`vtkDataSetReader (0x12aa7b20): Unable to open file: /home/travis/build/enthought/mayavi/mayavi/tests/data/UGridEx.vtk`

Other tests work fine.

Also, Travis uses VTK 5.8. I think if possible we should upgrade to VTK 5.10 on Travis and not bother about VTK < 5.10. 

This test works fine on VTK 5.10, i have tested locally. For now I have disabled this particular test (not worth spending time) on Travis to make it happy.

itziakos:
> Also, Travis uses VTK 5.8. I think if possible we should upgrade to VTK 5.10 on Travis and not bother about VTK < 5.10.

It is not easy. We will probably need to build VTK 5.10 ourselves, unless travis moved to using a more recent version of ubuntu.

itziakos:
> vtkDataSetReader (0x12aa7b20): Unable to open file: /home/travis/build/enthought/mayavi/mayavi/tests/data/UGridEx.vtk

It is possible that this file is not compatible with the reader in vtk 5.8. Does it pass locally with a vtk 5.8?

itziakos:
PR looks ok, we just need to check locally with 5.8 to be sure. Thanks for the fix :+1:

itziakos:
this PR is ready to merge

dmsurti:
Haven't yet tested with VTK 5.8 locally.

dmsurti:
Opened issue #219 for file reader unable to open unstructured grid file only on VTK 5.8.

Ready to merge.

Issue #216, opened 2015-07-14
Segmentation fault when using wx backend


dmsurti:
Using `ETS_TOOLKIT=wx`, wx Scene implementation can result in a seg fault.

The details of the backtrace:
`
Test the garbage collection of wx Scene implementation. ... 
(python2.7:27036): Gtk-CRITICAL **: IA__gtk_widget_set_double_buffered: assertion `GTK_IS_WIDGET (widget)' failed

(python2.7:27036): GLib-GObject-WARNING **: invalid (NULL) pointer instance

(python2.7:27036): GLib-GObject-CRITICAL **: g_signal_connect_data: assertion `G_TYPE_CHECK_INSTANCE (instance)' failed

(python2.7:27036): GLib-GObject-WARNING **: invalid (NULL) pointer instance

(python2.7:27036): GLib-GObject-CRITICAL **: g_signal_connect_data: assertion `G_TYPE_CHECK_INSTANCE (instance)' failed

(python2.7:27036): GLib-GObject-WARNING **: invalid (NULL) pointer instance

(python2.7:27036): GLib-GObject-CRITICAL **: g_signal_connect_data: assertion `G_TYPE_CHECK_INSTANCE (instance)' failed

(python2.7:27036): GLib-GObject-WARNING **: invalid (NULL) pointer instance

(python2.7:27036): GLib-GObject-CRITICAL **: g_signal_connect_data: assertion `G_TYPE_CHECK_INSTANCE (instance)' failed
/home/travis/build.sh: line 41: 27036 Segmentation fault      (core dumped) coverage run -p -m nose.core -v tvtk/tests
`

See the Travis CI Log: https://travis-ci.org/enthought/mayavi/jobs/70859123#L6998

itziakos:
I was able to replicate the problem without a segfault 

```python
In [1]: from tvtk.pyface.api import Scene

In [2]: a = Scene()
---------------------------------------------------------------------------
PyAssertionError                          Traceback (most recent call last)
<ipython-input-2-d749c40ddbf3> in <module>()
----> 1 a = Scene()

c:\users\itziakos\projects\mayavi\tvtk\pyface\ui\wx\scene.pyc in __init__(self, parent, **traits)
    301
    302         # Base class constructor.
--> 303         super(Scene, self).__init__(parent, **traits)
    304
    305         # Setup the default picker.

[...]

PR #218, opened 2015-07-15
Remove cycles from tvtk scenes

itziakos:
This PR updates #212 

In more detail:
- The test for the `TVTKScene` is reworked. 
- A separate tests is provided for the pyface based tvtk `Scene`. This test will use the default ets toolkit and currently will be skipped on `wx` (see #216)


tvtk/pyface/tvtk_scene.py
```diff
@@ -808,7 +816,7 @@ def _create_control(self, parent):
 
         self._interactor.initialize()
         self._interactor.render()
-        self.light_manager = light_manager.LightManager(self)
+        self.light_manager = light_manager.LightManager(weakref.proxy(self))
```
itziakos: @dmsurti, Please note that this fix removes the reference cycles which are collectible. There is no memory leak here. The cycle between `Scene` and `LightManager` will however keep the `Scene` alive if the `LightManager` is referenced by other components an application.

itziakos: Actually, it looks that it is not necessary to use a weakref here. Since as long as the `TVTKScene.close()` is called then the `_closed_fired` method will make sure that the cycle is properly broken.

dmsurti: Ok. I guess I must have used `pympler console browser`  to report back references which showed LightManager, before doing a `gc.collect`. If `_closed_fired` breaks the cycle, then not keeping it a weakref makes sense.

itziakos: Probably, but you where right that the cycle could cause problems

dmsurti:
LGTM. Thanks.

PR #220, opened 2015-07-16
Fixes #165: Fix Glyph to handle masking of input points

### LINKED ISSUE #165: change `mask_input_points` doesn't update the TVTK pipeline
Here is the code that set `mask_input_points` to True:

```
import numpy as np
from mayavi import mlab

p, r, b = (10.0, 28.0, 3.0)
x, y, z = np.mgrid[-17:20:20j, -21:28:20j, 0:48:20j]
u, v, w = p*(y-x), x*(r-z)-y, x*y-b*z

vectors = mlab.quiver3d(x, y, z, u, v, w)
vectors.glyph.mask_input_points = True
vectors.glyph.mask_points.on_ratio = 20
vectors.glyph.glyph.scale_factor = 5.0

mlab.show()
```

but the output is not masked. To fix this I need to change the following method in components/glyph.py:

```
    def _mask_input_points_changed(self, value):
        inputs = self.inputs
        if len(inputs) == 0:
            return
        if value:
            mask = self.mask_points
            tvtk_common.configure_input(mask, inputs[0].outputs[0])
            self.configure_connection(self.glyph, mask) # add this line
        else:
            self.configure_connection(self.glyph, inputs[0])
```

dmsurti:
This fixes #165. When mask input points are set, the glyph is now connected to the mask points filter.

Test Details: 
* [Added] to check the number of glyph input points when mask points filter random mode is set
* [Added] to check the number of glyph input points when mask points filter random mode is not set
* [Updated] existing test to check the number of glyph input points, without mask filter

For radom mode: see this: http://www.vtk.org/doc/nightly/html/classvtkMaskPoints.html#a21c7c2210111116f4b06ba9f020d0ed0

### DIFF
--- a/mayavi/components/glyph.py
+++ b/mayavi/components/glyph.py
@@ -271,8 +271,10 @@ class Glyph(Component):
         if value:
             mask = self.mask_points
             tvtk_common.configure_input(mask, inputs[0].outputs[0])
+            self.configure_connection(self.glyph, mask)
         else:
             self.configure_connection(self.glyph, inputs[0])
+        self.glyph.update()
 
     def _glyph_type_changed(self, value):
         if self.glyph_type == 'vector':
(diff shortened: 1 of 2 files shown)


prabhuramachandran:
This LGTM, the appveyor failure seems random and completely unrelated to this code.  I think we should attempt to fix that at a later point.  Merging this for now.

PR #222, opened 2015-07-17
Download the wxPython source from github

itziakos:
This PR fixes the problems with downloading the wx sources from sourceforge, by using the release files provided in github

PR #221, opened 2015-07-16
Fixes #178: Fix vtk cell insertion

dmsurti:
Fixes #178.

When a cell is inserted in VTK cell array, the number of cells is updated by ensuring `update_traits` is called.

Test added to check the number of cells in empty cell array and adding 2 cells in the cell array.

TVTK Wrapper generator `_write_tvtk_method` refactored to receive `klass` name as parameter, as `klass name` and `method name` are required to ensure `update_traits` call is added only for `cell_array.insert_next_cell`.

PR #223, opened 2015-07-17
Fixes #173: Allow long keys when looking up points

### LINKED ISSUE #173: tvtk.Points ``__getitem__`` does not accept python long integers
The following example demonstrates the issue

```
In [10] points = tvtk.Points()

In [11]: points.insert_next_point((0,1,2))
Out[11]: 0L

In [12]: points[0]
Out[12]: (0.0, 1.0, 2.0)

In [13]: points[0L]  # raises TypeError
```

dmsurti:
Fixes #173.

The special code generation in wrapping code is modified to allow look up with `long` keys in addition to existing int.

Test case added to test look up with `int`, `long` keys and also `get_point`.

---
Travis may fail for `qt4` with object_cache error, known issue and `wx` for wx src download problems. See PR #221 and #222.

### DIFF
--- a/tvtk/special_gen.py
+++ b/tvtk/special_gen.py
@@ -319,8 +319,11 @@ class SpecialGenerator:
                 yield obj.GetPoint(i)
 
         def _check_key(self, key, n):
-            if type(key) != type(1):
-                raise TypeError, "Only integers are valid keys."
+            ##############################################
+            # Allow int and long keys. Fixes GH Issue 173.
+            ##############################################
+            if type(key) not in [type(1), type(1L)]:
+                raise TypeError, "Only int and long are valid keys."
             if key < 0:
                 key =  n + key
             if key < 0 or key >= n:
--- a/tvtk/special_gen.py
+++ b/tvtk/special_gen.py
@@ -322,7 +322,7 @@ class SpecialGenerator:
             ##############################################
             # Allow int and long keys. Fixes GH Issue 173.
             ##############################################
-            if type(key) not in [type(1), type(1L)]:
+            if not isinstance(key, (int, long)):
                 raise TypeError, "Only int and long are valid keys."
             if key < 0:
                 key =  n + key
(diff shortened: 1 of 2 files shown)


tvtk/special_gen.py
```diff
@@ -319,8 +319,11 @@ def __iter__(self):
                 yield obj.GetPoint(i)
 
         def _check_key(self, key, n):
-            if type(key) != type(1):
-                raise TypeError, "Only integers are valid keys."
+            ##############################################
+            # Allow int and long keys. Fixes GH Issue 173.
+            ##############################################
+            if type(key) not in [type(1), type(1L)]:
```
itziakos: I do not think it is necessary to check for type. We should just propagate the key in this case and let the vtk python to coerce the value and fail (if necessary).

itziakos: If key is not a numeric value it will fail anyway in the comparisons that follow.

dmsurti: @prabhuramachandran Can you tell why the `_check_key` calls have been added? Was that necessary for VTK < 5.10? Is it safe to drop these calls, as they are also used for IdList and DataArray?

prabhuramachandran: IIRC, the main reason for `_check_key` is to support negative indices and raise the correct exceptions where necessary.  VTK will not raise any exceptions for those cases, which could lead to a segfault.  Try this:  
```
import vtk
d = vtk.vtkDoubleArray()
d.SetNumberOfTuples(5)
d.GetTuple(-1)
d.GetTuple(100)
```
You might get a segfault!  The integer/long checks are not technically necessary but should not be harmful.  VTK will raise the exception but the user may be confused when they see the exception so I think I caught it earlier.

dmsurti: So is it fine to retain the `_check_key` call?

itziakos: keeping the `_check_key` call is ok for know, but we should rework the method to only check the for negative numbers and not the type anymore.

itziakos: probably use `isinstance` here

itziakos:
:+1:

PR #230, opened 2015-07-23
Clean up memory leaks implementation in #218

dmsurti:
This is a clean up for PR #218.

1. Remove `psutil` dependency. Moved the `MemoryAssistant` from `tvtk.common` to `integrationtests.common`.
2. Conditional import of `psutil`.
3. Call super `_closed_fired` for `wx Scene` implementation.

### DIFF
--- a/integrationtests/mayavi/common.py
+++ b/integrationtests/mayavi/common.py
@@ -5,6 +5,7 @@
 # License: BSD Style.
 
 # Standard library imports
+import gc
 import os
 import os.path
 import sys
@@ -17,7 +18,6 @@ from traits.etsconfig.api import ETSConfig
 from traits.api import  Any, Bool, Instance
 from pyface.api import GUI
 from tvtk.api import tvtk
-from tvtk.common import MemoryAssistant
 from mayavi.plugins.app import Mayavi, setup_logger
 
 # The TVTK window.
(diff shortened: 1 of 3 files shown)


integrationtests/mayavi/common.py
```diff
@@ -461,6 +573,13 @@ def parse_command_line(self, argv):
         if options.verbose:
             VERBOSE = True
             self.log_mode = logging.DEBUG
+        global psutil
```
prabhuramachandran: Umm this is dirty, instead just perform the import in each function that requires it.  i.e. above line 106.

dmsurti: Done. 962b4a4fd83217f69a1a9e0da140ed97cfb26084

prabhuramachandran:
LGTM! :+1:

tvtk/pyface/ui/wx/scene.py
```diff
@@ -569,6 +569,7 @@ def OnButtonUp(self, event):
     ###########################################################################
 
     def _closed_fired(self):
+        super(Scene, self)._closed_fired()
         self.picker = None
         self.light_manager = None
```
itziakos: the `parent` method will set 
```python
        self.light_manager = None
        self._interactor = None
```
no need to repeat the same operations

itziakos:
one minor comment, other than that it looks good to me.

PR #231, opened 2015-07-23
Remove ref cycles in MlabSceneModel, Core Scene

dmsurti:
This is a follow up to PR #218 for memory leaks.

---
Overall the API is evolving as such: The Scene API has a closed event handler which should be invoked from the client app. When a scene is closed, the engine closed handler is also invoked. Also for example: if an app uses `MlabSceneModel` then setting `closed = True` in the app's appropriate handler (such as when UI is destroyed) should ensure that scene objects are garbage collected.

Final Update
---
Added tests for `DecoratedScene, Mayavi UI Scene` and minor refactoring: invoke close() fn for scenes and close engine when it is stopped.

Update 4
---
This removes ref cycles in `mayavi.core.Engine` and
1. Scenes it manages
2. Root preferences helper whose `show_helper_nodes` trait referred back to Engine via a `lambda`.

The engine now has a `closed` event handler which is invoked when the scene is closed.

Update 3:
---
This removes the ref cycles in `mayavi.core.Scene` and:
1. Mayavi core engine
2. Mouse pick dispatcher which has a `__del__`.

A closed event handler has been added to scene which is also now invoked from MlabSceneModel.

Update 2:
---
Refactored the unit test design for garbage collection. There is a base class TestGarbageCollection in `tvtk.tests.common` that has a method `check_object_garbage_collected` which does all the housekeeping. A test method invokes this `check_` method passing a function for object creation and another optional function to handling closing of the object, if any, where the object is the one to be tested for gc.

There are 2 test_garbage_collection.py modules, one each for `tvtk` and `mayavi` packages. These now cover garbage collection of Scene, SceneModel and MlabSceneModel objects.

Update 1:
----
This removes cycles between MlabSceneModel and:
1. Mayavi Core Scene
2. Mayavi Core Engine

These cycles are cleared in the `closed` event of the scene model.

A unit test case is added to test the reference cycles.

As MlabSceneModel is a subclass of SceneModel, that has also been tested for reference cycles. SceneModel also has `closed` event handling now.

Minor addition to scene ref cycle test case for testing weakref as well after gc,


dmsurti:
Again the poly data reader test fails, unrelated to these code changes. See Issue #233.

mayavi/core/registry.py
```diff
@@ -72,7 +73,7 @@ def unregister_engine(self, engine_or_name):
                     name = key
                     break
 
-        del engines[name]
+        if name: del engines[name]
```
itziakos: I would suggest to not have multiple statements in a single line

mayavi/tests/test_mlab_scene_model.py
```diff
@@ -1,11 +1,15 @@
 """
 Testing the MlabSceneModel
 """
+import gc
```
itziakos: It looks that these imports where added but not used

tvtk/pyface/scene_model.py
```diff
@@ -328,6 +333,9 @@ def _get_light_manager(self):
     ######################################################################
     # SceneModel API.
     ######################################################################
+    def _closed_fired(self):
```
itziakos: I do not think there is a need to do this if all you are doing is calling the parent class

tvtk/pyface/scene_model.py
```diff
@@ -57,9 +59,12 @@ class SceneModel(TVTKScene):
     # Fired when this is activated.
     activated = Event()
 
-    # Fired when this widget is closed.
+    # The control is going to be closed.
     closing = Event()
 
+    # The control has been closed.
```
itziakos: TVTKScene already defines these traits do we need to redefine them?

tvtk/pyface/ui/qt4/decorated_scene.py
```diff
@@ -71,6 +71,9 @@ def __get_pure_state__(self):
     ##########################################################################
     # Non-public interface.
     ##########################################################################
+    def _closed_fired(self):
```
itziakos: Again, I do not think this is necessary

tvtk/tests/common.py
```diff
@@ -12,3 +20,42 @@ def restore_gc_state():
...
+        object_collected = []
+        object_weakref = None
+
+        def object_collected_callback(weakref):
+            object_collected.append(True)
+
+        def do():
+            obj = obj_create_fn()
+            reference = weakref.ref(obj, object_collected_callback)
+            if obj_close_fn: obj_close_fn(obj)
```
itziakos: do not place multiple statements in one line.

mayavi/core/engine.py
```diff
@@ -567,6 +573,19 @@ def _on_scene_activated(self, obj, name, old, new):
...
                 break
 
+    def _closed_fired(self):
+        """ When the scene is closed, clear the viewer ref which otherwise
+            stores references to scenes to prevent crash on QT4.
+            See: self.new_scene and MlabSceneModel._closed_fired
+        """
+        self._viewer_ref.clear()
+        self.scenes = []
+        registry.unregister_engine(self)
```
itziakos: we should make sure that the dynamic change handler is removed 
```python
preference_manager.root.on_trait_change(self._show_helper_nodes_changed,
                                                 'show_helper_nodes', remove=True)
```

mayavi/core/engine.py
```diff
@@ -567,6 +573,22 @@ def _on_scene_activated(self, obj, name, old, new):
                 self.current_scene = scene
                 break
 
+    def _closed_fired(self):
+        """ When the scene is closed, clear the viewer ref which otherwise
```
prabhuramachandran: When the "engine" is closed and not scene as written.

mayavi/core/engine.py
```diff
@@ -101,6 +101,10 @@ class Engine(HasStrictTraits):
     # Are we running?
     running = Bool(False, record=False)
 
+    # The engine has been closed. This does not stop running the engine.
```
prabhuramachandran: Instead you could comment as: "This event is invoked when the engine has been stopped" and don't confuse this with the scenes being closed as that is entirely different and will confuse the reader.

mayavi/core/scene.py
```diff
@@ -198,6 +200,11 @@ def tno_drop_object(self, node, dropped_object):
     ######################################################################
     # Non-public interface
     ######################################################################
+    def _closed_fired(self):
+        self.scene = None
+        self.menu_helper.object = None
+        self.parent.closed = True
```
prabhuramachandran: This should really never call engine's closed as that will remove/clear all viewer refs etc. in the engine.  Instead engine's remove_scene method could be called.  However, this is tricky as if the TVTK Scene plugin is used, it automatically takes care of removing the scenes from the engine.  I think this new event "closed" is not needed.  Let us discuss this.

prabhuramachandran: Basically, the `SceneModel` and other subclasses have a "closing" event that is automatically hooked up to close references in `engine.new_scene`.

mayavi/tests/test_garbage_collection.py
```diff
@@ -0,0 +1,74 @@
...
+from mayavi.core.ui.mayavi_scene import MayaviScene
+from mayavi.tools.mlab_scene_model import MlabSceneModel
+
+class TestMayaviGarbageCollection(TestGarbageCollection):
+    """ See: tvtk.tests.common.TestGarbageCollection
+    """
+    def test_mlab_scene_model(self):
+        """ Tests if MlabSceneModel can be garbage collected."""
+        def create_fn():
+            return MlabSceneModel()
```
prabhuramachandran: This will not really test much, it will be better if you create a class which has an MlabSceneModel traits, configure the traits (so the UI pops up), close it, and then ensure that everything is GC'd.  In particular, the engine's  _viewer_refs should be empty.

dmsurti: I am not sure if MlabSceneModel really needs to be tested with a traits UI. The MlabSceneModel when initialized already registers itself with the engine. So if I set `closing=True` and it is garbage collected, that implies the engine's _viewer_refs are empty, else it won't be GC'ed. Isn't that right?

itziakos: I think that we need both tests types but the gui test is probably more important. One can use the `invoke_after` method of the GUI class to call a function that will close the window.

mayavi/core/scene.py
```diff
@@ -58,6 +58,9 @@ class Scene(Base):
         'mayavi.core.mouse_pick_dispatcher.MousePickDispatcher',
         record=False)
 
+    # The scene has been closed.
+    closed = Event()
```
itziakos: It looks that this event is not used anywhere. It should be removed then to avoid confusion

itziakos:
There are a couple of docstring comments pending apart from that it looks ok

mayavi/tests/test_garbage_collection.py
```diff
@@ -0,0 +1,110 @@
...
+                X, Y = mgrid[-2:2:100j, -2:2:100j]
+                R = 10*sqrt(X**2 + Y**2)
+                Z = sin(R)/R
+                self.scene.mlab.surf(X, Y, Z, colormap='gist_earth')
+
+        def create_fn():
+            app = MlabApp()
+            return app.edit_traits()
+
+        self.check_object_garbage_collected(create_fn)
```
itziakos: I think that you need to use another dummy class with `create and destroy` methods.

```python
class Dummy(object):

    def create(self):
          self.app = app = MlabApp()
          self.ui = app.edit_traits()
          return app.scene

    def destroy(self):
          self.ui.dispose()
          self.ui = None
          self.app = None
```

To make sure that dispose is called

itziakos:
I think that this PR can be merged :+1:

PR #235, opened 2015-07-29
BUG: Selection of LUT broken with qt backend.

### LINKED ISSUE #147: On large hi-res screen on Mac, LUT color picker mis-identifies palette selection
User report. On MacPro with two AMD FirePro D500 graphics cards and two Thunderbolt displays with 2560 x 1440 resolution. Scalar LUT manager chosen, 256 colors, 60 palettes shown (plus one "binary file"). In most cases when a palette is clicked, a different palette (usually but not always to the right of the one clicked) is actually chosen and used. Video available.

prabhuramachandran:
With the qt backend it is impossible to select the right `lut_mode`.
This fixes the problem and fixes #147.

### DIFF
--- a/mayavi/core/ui/lut_manager.py
+++ b/mayavi/core/ui/lut_manager.py
@@ -18,15 +18,13 @@ from traitsui.api \
 from mayavi.core.lut_manager import lut_mode_list, \
             lut_image_dir
 
-if ETSConfig.toolkit == 'qt4':
-    lut_cols = 1
-else:
-    lut_cols = 6
+def _number_of_lut_cols():
+    return 1 if ETSConfig.toolkit == 'qt4' else 6
 
 # The view of the LUT Manager object.
 view = View(Group(Item(name='lut_mode',
                        editor=ImageEnumEditor(values=lut_mode_list(),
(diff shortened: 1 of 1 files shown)


mayavi/core/ui/lut_manager.py
```diff
@@ -9,19 +9,24 @@
...
+# Copyright (c) 2005-2015, Enthought, Inc.
 # License: BSD Style.
 
+from traits.etsconfig.api import ETSConfig
 from traitsui.api \
     import Item, Group, View, ImageEnumEditor, InstanceEditor, HGroup
 from mayavi.core.lut_manager import lut_mode_list, \
             lut_image_dir
 
+if ETSConfig.toolkit == 'qt4':
```
dmsurti: Can't this be `lut_cols()` similar to `lut_mode_list()`. Makes it consistent.

dmsurti: With lut_cols as a function, writing a test should also be easy I guess.

prabhuramachandran:
Updated as per comments.  Thanks.

PR #237, opened 2015-08-10
Remove redefined trait from scene model 

dmsurti:
Based on @prabhuramachandran  comment for PR #231, remove redefined light manager trait from scene model, which is not necessary. This was added when `closed` event was used to remove ref cycles.


itziakos:
:+1:

prabhuramachandran:
Thanks, LGTM. :+1:

PR #238, opened 2015-08-11
Changes for the 4.4.3 release

dmsurti:
1. Bump up the version number.
2. Update the change log.

### DIFF
--- a/docs/CHANGES.txt
+++ b/docs/CHANGES.txt
@@ -40,9 +40,6 @@ Fixes
     - Fix TVTK wrapper to update number of cells when a cell is inserted in a
       VTK cell array
 
-20 Jul 2015 (IT)
-    - Download wx sources for CI from release files on github
-
 15 Jul 2015 (DS)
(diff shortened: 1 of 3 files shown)


docs/CHANGES.txt
```diff
@@ -1,3 +1,54 @@
...
+
+22 Jul 2015 (IT)
+    - Update the docs with known bugs, issues list and describe the PR
+      workflow on github for development and documentation.
+
+20 Jul 2015 (DS)
+    - Fix TVTK wrapper to update number of cells when a cell is inserted in a
+      VTK cell array
+
+20 Jul 2015 (IT)
```
itziakos: This change is obsolete since we now use wx-tools in travis-ci, I do not think we need to add this update.

itziakos:
On small comment apart from that :+1:

prabhuramachandran:
Yes, agree with @itziakos , otherwise LGTM.

prabhuramachandran:
:+1:

PR #239, opened 2015-08-11
Updated docs for 4.4.3 release

dmsurti:
The updated docs for the next Mayavi release 4.4.3.


itziakos:
PR looks ok. 

The only strange thing is that gh-pages contains a copy of the actuall mayavi source code which as far as I know is uncommon for gh-pages (see https://github.com/enthought/chaco/tree/gh-pages), but this can be discussed offline and does not block the current release.

dmsurti:
This is ready for merge.

PR #249, opened 2015-09-08
Ensure cutter is updated when cut plane is changed.

### LINKED ISSUE #164: cut plane doesn't update when the plane is changed by GUI
Here is the code:

```
import numpy as np
from mayavi import mlab

field = mlab.pipeline.scalar_field(s)
mlab.pipeline.volume(field, vmin=1.5, vmax=10)
cut = mlab.pipeline.scalar_cut_plane(field.children[0], plane_orientation="y_axes")
cut.enable_contours = True
cut.contour.number_of_contours = 40
mlab.gcf().scene.background = (0.8, 0.8, 0.8)
mlab.show()
```
cut_plane

To fix this I need to add following code to the `Cutter` class in components/cutter.py:

```
def _cut_function_changed(self):
    self.cutter.cut_function.on_trait_change(self.cutter.update, "normal, origin")
```

dmsurti:
Fixes #164.

1. The implicit plane fires `pipeline_changed` which ensures `cutter` is updated.
2. Unit test to test with both structured and unstructured data that cutter output changes when cut plane normal/origin is changed.
3. Integration test.

### DIFF
--- a/mayavi/components/implicit_plane.py
+++ b/mayavi/components/implicit_plane.py
@@ -149,15 +149,18 @@ class ImplicitPlane(Component):
     ######################################################################
     def _get_normal(self):
         return self.widget.normal
+
     def _set_normal(self, value):
         w = self.widget
         old = w.normal
         w.normal = value
         self.trait_property_changed('normal', old, value)
         self.update_plane()
+        self.pipeline_changed = True
 
     def _get_origin(self):
         return self.widget.origin
+
     def _set_origin(self, value):
         # Ugly, but needed.
         w = tvtk.to_vtk(self.widget)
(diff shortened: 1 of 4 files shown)


mayavi/components/implicit_plane.py
```diff
@@ -149,15 +149,18 @@ def update_plane(self):
...
     def _get_normal(self):
         return self.widget.normal
+
     def _set_normal(self, value):
         w = self.widget
         old = w.normal
         w.normal = value
         self.trait_property_changed('normal', old, value)
         self.update_plane()
+        self.pipeline_changed = True
```
prabhuramachandran: Now that you have some tests, it may be good to see if there is a less performance intense approach to fix this.  A pipeline change is costly as it flushes everything downstream, i.e. all connections are re-made.  A data_changed is cheaper but may not do the trick, perhaps an explicit update call somewhere may help, I am not sure but `pipeline_changed` is costly.

prabhuramachandran:
A better fix is being worked on and a PR will be sent soon, closing this for now.

PR #255, opened 2015-10-13
Bug #163 contour3d() shows only one contour with contours = N

### LINKED ISSUE #163: contour3d() shows only one contour with contours = N
Here is the code

```
import numpy as np
from mayavi import mlab


surface = mlab.contour3d(s, contours=4, transparent=True)

mlab.show()
```
There is only one contour in the scene:
contours

I checked the code and can fix the problem by modify the following function in mayavi/components/contour.py:

```
    def _do_auto_contours(self):
        if not self._has_input():
            return
        if self.auto_contours:
            minc, maxc = self.minimum_contour, self.maximum_contour
            self.contour_filter.generate_values(self.number_of_contours,
                                                min(minc, maxc),
                                                max(minc, maxc))
            self.contour_filter.update() #add this line
            self.data_changed = True
```

maheshpj:
Fixes #163 
1. When auto contour is True make sure contour filter is getting updated
2. Update contour unit test to check number of contour points before and after auto_contour is True
3. Update contour unit test to check contour values are within min_contour and max_contour values


dmsurti:
Closed by recent pipeline changes, #309.

PR #247, opened 2015-09-01
Update to VTK Version 6.3

dmsurti:
This PR tests Mayavi compatibility with VTK 6.3 official release.

PR #250, opened 2015-09-08
Python3 support

prabhuramachandran:
The aim of this PR is to solicit feedback and merge this early for those who want to cut themselves on the bleeding edge.  

This adds experimental Python3 support for Mayavi.  To use it one requires a recent version of apptools and envisage (by recent I mean that these PRs should be merged https://github.com/enthought/apptools/pull/54 and https://github.com/enthought/envisage/pull/61 ).

It requires the master branch of pyface. 

It will require the latest version of VTK from their git repository (https://gitlab.kitware.com/vtk/vtk) and a fresh build of VTK configured for Python3.  I have not bothered with travis CI support at this point due to the complexity of doing that.  These changes pass all current 2.7.x tests on travis so should be safe to merge.

## Status and known issues

 * The changes are compatible with Python 2.x and 3.x.  I have tested on 2.7.x and 3.4.x.

 * All the tvtk tests pass.

 * There are a couple of random failures in the mayavi tests.  These may be version specific issues but need more investigation.

 * The integration tests mostly run fine but blocks on exit (with PyQt) so if you run one of them you need to kill it to run the next.  This can be investigated later.

 * mlab works quite well.

 * The mayavi2 application itself has a few issues that seem pyface related and will be investigated later, notably the persistence of the workbench perspectives are not pickled/unpickled properly..

PR #269, opened 2015-12-08
Check for a `find` method of parsed VTK docstring

### LINKED ISSUE #261: Error in vtk_parses while installation on Mac OS X 10.9.5
HI,

I get the following error:
File "tvtk/vtk_parser.py", line 311, in get_method_signature
AttributeError: 'NoneType' object has no attribute 'find'

This happens both when I install from source or use macports. Easy-install also doens't work for me. Is there a way to fix it?

fish2000:
The VTK parser method `get_method_signature` now uses `hasattr()` to confirm that the `__doc__` attributes it is inspecting do in fact have valid string methods, prior to calling any of those methods.

This fixes #261 – a bug that, without this change, will persist for me when building mayavi on OS X (10.10 and 10.11) regardless of whether I apply PR #247 (q.v. http://git.io/vRS8E).

### DIFF
--- a/tvtk/vtk_parser.py
+++ b/tvtk/vtk_parser.py
@@ -308,6 +308,8 @@ class VTKMethodParser:
                 return None
         # Remove all the C++ function signatures.
         doc = method.__doc__
+        if not hasattr(doc, 'find'):
+            return None
         doc = doc[:doc.find('\n\n')]
         sig = []
         c_sig = [] # The C++ signature


prabhuramachandran:
I suspect this is no longer needed with master as I think I have fixed some of these issues in #250 that is now merged.  Please re-open if this is still an issue.

prabhuramachandran:
@fish2000 -- sorry about my lack of manners and thank you for the original PR!

fish2000:
@prabhuramachandran oh hey no problem, my pleasure – you are correct, the #250 merge did the trick. Thanks!

PR #278, opened 2016-01-21
Do not use get-pip or wheels when upgrading pip

itziakos:


dmsurti:
@itziakos LGTM. Thanks.

PR #284, opened 2016-02-04
make sure that the __file__ points to the srcipt file when running scripts

itziakos:
I recently faced a problem were running a script from Mayavi2 will mess-up the `__file__` attribute of the script. Using the `__file__` attribute is common for loading data so having the atrribute point to the right path is important.

The change is small and looks harmless, but I am not completely sure. I would also like ideas on how to test this.

@dmsurti  and @prabhuramachandran  please have a look


prabhuramachandran:
This seems fine to me.  Not sure this is used much at all so not sure it is worth any testing effort.  I am fine merging this as is and will do so.  Thanks.

Issue #281, opened 2016-01-29
Traceback when interacting with scene, keeping a key pressed


pberkes:
On a slower system, like a Linux Virtual Machine, I observed an occasional traceback when pressing a key (say, `z`) and rotating the scene with the mouse:
```
Traceback (most recent call last):
  File "/home/berkes/Enthought/Canopy_64bit/System/lib/python2.7/site-packages/tvtk/pyface/ui/qt4/scene.py", line 198, in keyPressEvent
    QVTKRenderWindowInteractor.keyPressEvent(self, e)
  File "/home/berkes/Enthought/Canopy_64bit/System/lib/python2.7/site-packages/tvtk/pyface/ui/qt4/QVTKRenderWindowInteractor.py", line 418, in keyPressEvent
    ctrl, shift, key, 0, key_sym)
TypeError: SetEventInformationFlipY argument 5: a string of length 1 is required
```

The error is caused by the `key` argument sent to `SetEventInformationFlipY` being something like `"zzzzz"`, i.e. multiple repetitions of the key I am pressing.

Using advanced debugging techniques, I edited the `keyPressEvent` method of `tvtk.pyface.ui.qt4.QVTKRenderWindowInteractor` like this:

```
    def keyPressEvent(self, ev):
        ctrl, shift = self._GetCtrlShift(ev)
        key_sym = self._KEY_MAP.get(ev.key(), None)
        if ev.key() < 256:
            # Sometimes, the OS allows a chord (e.g. Alt-T) to generate
            # a Unicode character outside of the 8-bit Latin-1 range. We will
            # try to pass along Latin-1 characters unchanged, since VTK expects
            # a single `char` byte. If not, we will try to pass on the root key
            # of the chord (e.g. 'T' above).
            if ev.text() and ev.text() <= u'\u00ff':
		print '1'
                key = ev.text().encode('latin-1')
            else:
		print '2'
                # Has modifiers, but an ASCII key code.
                key = chr(ev.key())
        else:
	    print '3'
            key = chr(0)

	print '-', self.__saveX, self.__saveY, ctrl, shift, key, 0, key_sym, '-'
        self._Iren.SetEventInformationFlipY(self.__saveX, self.__saveY,
                                            ctrl, shift, key, 0, key_sym)
        self._Iren.KeyPressEvent()
        self._Iren.CharEvent()
```

While interacting with the scene, this is what I get on stdout:

```
[...]
1
- 241 465 False False z 0 None -
1
[...]

PR #286, opened 2016-02-08
Fix to avoid multiple identical keys in keyPressEvent

jonathanrocher:
In the Qt implementation of tvtk's `QVTKRenderWindowInteractor`, I propose to truncate the key pressed to first character. This has been tested to avoid exceptions thrown by `SetEventInformationFlipY` if slow machine leads to multiple times the same key in a given event.

Fixes https://github.com/enthought/mayavi/issues/281


jonathanrocher:
@dmsurti As discussed in the issue thread...

dmsurti:
Merged.

Issue #197, opened 2015-06-12
`to_array` raises ValueError when the vtk array was recently appended


itziakos:
The following script will raise a nice ValueError

```python
from tvtk.api import tvtk
import numpy

data = tvtk.PointData()
data.add_array(numpy.zeros(shape=(0, 3)))
temp = data.get_array(0)
print temp.to_array(), temp.to_array().shape
temp.append([0, 0, 0.4])
temp.to_array()
```

itziakos:
traceback
```python
ValueError                                Traceback (most recent call last)
c:\users\itziakos\appdata\local\temp\tmps1jdlk.py in <module>()
      7 print temp.to_array(), temp.to_array().shape
      8 temp.append([0, 0, 0.4])
----> 9 temp.to_array()
     10 
     11 

C:\Users\itziakos\AppData\Local\Enthought\Canopy\User\lib\site-packages\tvtk\tvtk_classes.zip\tvtk_classes\data_array.pyc in to_array(self)
    711     def to_array(self):
    712         '''Return the object as a Numeric array.'''
--> 713         return array_handler.vtk2array(self._vtk_obj)
    714 
    715 

C:\Users\itziakos\AppData\Local\Enthought\Canopy\User\lib\site-packages\tvtk\array_handler.py in vtk2array(vtk_array)
    353         if shape[1] == 1:
    354             shape = (shape[0], )
--> 355         arr = numpy.reshape(arr, shape)
    356         return arr
    357 

C:\Users\itziakos\AppData\Local\Enthought\Canopy\User\lib\site-packages\numpy\core\fromnumeric.pyc in reshape(a, newshape, order)
    219     except AttributeError:
    220         return _wrapit(a, 'reshape', newshape, order=order)
--> 221     return reshape(newshape, order=order)
    222 
    223 

ValueError: total size of new array must be unchanged 
```

itziakos:
currently this `_array_cache._remove_array(tvtk.to_vtk(temp).__this__)` will invalidate the cache for the vtk array.

itziakos:
the problem is that when the `append` takes place in the `vtkDoubleArray` the `_array_cache` of `mayavi` is not invalidated. As a result the `to_array` method will get the old numpy array that is still in the cache and try to reshape it to the new shape of the vtkDoubleArray which correctly fails. One solution for this case is to make sure that the cache is invalidated when the `append` method is called on a vtk array.

PR #289, opened 2016-02-15
make sure the shape matches the size of array in `to_array`

kitchoi:
This is for issue #197 
@prabhuramachandran, @dmsurti  can you take a look please?

### DIFF
--- a/tvtk/tests/test_array_handler.py
+++ b/tvtk/tests/test_array_handler.py
@@ -396,6 +396,20 @@ class TestArrayHandler(unittest.TestCase):
         del varr
         self.assertEqual(len(cache), 0)
 
+    def test_vtk2array_appended_array(self):
+        """Test the vtk2array can tolerate appending a cached array."""
+        # array is cached upon array2vtk is called
+        arr  = numpy.arange(8).reshape(2, 4)
+        vtk_arr = array_handler.array2vtk(arr)
+        arr1 = array_handler.vtk2array(vtk_arr)
+
+        # the vtk array is appended, shapes don't match cached array anymore
+        vtk_arr.InsertTuple4(2, 1., 2., 3., 4.)
+
+        # arr2 has a different shape
+        arr2 = array_handler.vtk2array(vtk_arr)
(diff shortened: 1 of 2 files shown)


dmsurti:
@kitchoi Can you add a test in `tvtk/tests/test_array_handler.py`?

kitchoi:
@dmsurti Sure.

kitchoi:
@dmsurti I am wondering if the cache should be cleared in the same place where I added the shape check.  The cache is there to prevent the user from modifying the array being sent to the vtk, but this PR is allowing the opposite.  In other words, is the cache still needed?  (i can't find `array_cache.get` anywhere else)

dmsurti:
> The cache is there to prevent the user from modifying the array being sent to the vtk, but this PR is allowing the opposite

Nope. The cache is to prevent only from deleting/resizing. https://github.com/enthought/mayavi/blob/master/tvtk/array_handler.py#L61.

tvtk/tests/test_array_handler.py
```diff
@@ -396,6 +396,20 @@ def test_array_cache(self):
...
+        arr  = numpy.arange(8).reshape(2, 4)
+        vtk_arr = array_handler.array2vtk(arr)
+        arr1 = array_handler.vtk2array(vtk_arr)
+
+        # the vtk array is appended, shapes don't match cached array anymore
+        vtk_arr.InsertTuple4(2, 1., 2., 3., 4.)
+
+        # arr2 has a different shape
+        arr2 = array_handler.vtk2array(vtk_arr)
+        self.assertEqual(arr2.shape, (3, 4))
```
itziakos: you should also check the values of the array to make sure that all is as expected

kitchoi: Good point.  Should inserting float values to a vtk array originally initialised with int values trigger a type casting?
(Separate question, what is the lowest version requirement of numpy for mayavi?)

prabhuramachandran:
@kitchoi -- thanks for catching this!  LGTM.

PR #293, opened 2016-02-19
Update doc setup

kitchoi:
Previous build of documentation may have relied on cached files.  I need to make these changes in order to build the documentation properly.  
@dmsurti , @prabhuramachandran Can you verify this please?
I am using Sphinx 1.3.5


docs/source/mayavi/conf.py
```diff
@@ -23,7 +23,7 @@
 
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
-extensions = ['sphinx.ext.autodoc', 'traitsdoc']
+extensions = ['sphinx.ext.autodoc', 'sphinxext.traitsdoc']
```
prabhuramachandran: Is this needed?  The first line adds the sphinxext directory to the path, so shouldn't this just work?

docs/source/mayavi/sphinxext/sphinxext/comment_eater.py
```diff
@@ -1,10 +1,7 @@
 from __future__ import division, absolute_import, print_function
 
 import sys
-if sys.version_info[0] >= 3:
-    from io import StringIO
-else:
-    from io import StringIO
+from io import StringIO
```
prabhuramachandran: These files are copied over from http://github.com/numpy/numpydoc
not sure what is wrong.

prabhuramachandran:
Hmm, so yes `make html` does not work on master.  `python setup.py build_docs` also does not work on master and everything works with your changes.  I am not sure why you needed to move the sphinxext into another nested sphinxext?  Also, this does not work on Python3 at all but that has nothing to do with your PR.

prabhuramachandran:
OK, your changes actually look good and fix the problems on Python 2.7 so am merging this for now.  The Python3 issues can be fixed later.  Thanks!

PR #296, opened 2016-02-22
Include OffScreenEngine in get_engine

kitchoi:
This PR is proposing an enhancement so that when the `mlab.options.offscreen` is True and `options.backend` is not "envisage" or "test", the `mlab.get_engine()` function would return an `OffScreenEngine`.

This will remove the need to switch on `mlab.options.offscreen` before instantiating any `Engine`.

I haven't proceeded with test cases yet.  I would to know whether this is a good idea or not.


kitchoi:
@prabhuramachandran , @dmsurti  what do you think?

prabhuramachandran:
I think this is a good idea but I think it does need a round of testing (at least manual), one with the mayavi2 application and one with traitsui dialogs on an IPython console.

kitchoi:
@prabhuramachandran  I have added a test case for this PR, introduced `mock` to travis-requirement.  The test case should be fairly self-explanatory which engine is returned under what situation.  Does this look reasonable to you?

prabhuramachandran:
LGTM, thanks, the tests look very nice!

PR #297, opened 2016-02-25
Fix QVTKRenderWindowInteractor._get_win_id()

mabl:
Under Python 3 and Windows, the conversion str(WId) will return
the representative string of the object instead of the integer id
as string. This fix mirrors the behaviour of VTK's own
QVTKRenderWindowInteractor.

For more information on this bug, see
 http://vtk.1045678.n5.nabble.com/OpenGL-fails-to-initialize-for-Windows-7-VTK-7-0-td5736723.html


prabhuramachandran:
Good find!  LGTM!  Thanks!

prabhuramachandran:
Thanks for hunting this down!  LGTM! Merging.

PR #298, opened 2016-02-26
Fix cutter

### LINKED ISSUE #164: cut plane doesn't update when the plane is changed by GUI

prabhuramachandran:
Fixes #164.  There are potentially more pipeline changes to make but will be done in a separate PR.

### DIFF
--- a/mayavi/components/actor.py
+++ b/mayavi/components/actor.py
@@ -2,7 +2,7 @@
 
 """
 # Author: Prabhu Ramachandran <prabhu_r@users.sf.net>
-# Copyright (c) 2005, Enthought, Inc.
+# Copyright (c) 2005-2016, Enthought, Inc.
 # License: BSD Style.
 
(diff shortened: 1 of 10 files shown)


dmsurti:
:+1: Merged.

Issue #288, opened 2016-02-09
Failures with VTK 7.0


dmsurti:
I tested with VTK 7.0 on Mac OS X 10.10. These are the failures: 
* TVTK:   FAILED (errors=1, failures=5)
* Mayavi: FAILED (errors=1).

TVTK Failures
---
```
...........F......SS...........................F....S.......E.......F..............FF......
======================================================================
ERROR: Case where parent has GetInput and child SetInput.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/deepaksurti/repos/git/mayavi/tvtk/tests/test_tvtk.py", line 599,
in test_parent_child_input
    vm = tvtk.VolumeTextureMapper2D()
AttributeError: 'TVTK' object has no attribute 'VolumeTextureMapper2D'

======================================================================
FAIL: Basic tests for the VTK module.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/deepaksurti/repos/git/mayavi/tvtk/tests/test_class_tree.py",
line 79, in test_basic_vtk
    self.assertEqual(names, expect)
AssertionError: Lists differ: ['object', 'vtkColor3', 'vtkCo... != ['object',
'vtkColor3', 'vtkCo...

First differing element 9:
vtkVariantStrictWeakOrderKey
vtkVector

First list contains 1 additional elements.
First extra element 12:
vtkVector3

  ['object',
   'vtkColor3',
   'vtkColor4',
   'vtkDenseArray',
   'vtkQuaternion',
   'vtkRect',
   'vtkSparseArray',
   'vtkTuple',
   'vtkTypedArray',
-  'vtkVariantStrictWeakOrderKey',
   'vtkVector',
   'vtkVector2',
   'vtkVector3']

======================================================================
FAIL: Test if custom modules can be imported.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/deepaksurti/repos/git/mayavi/tvtk/tests/test_tvtk.py", line 114,
in test_custom
    self.assertEqual(r.__class__.__bases__, expect)
AssertionError: Tuples differ: () != (<type 'object'>,)

Second tuple contains 1 additional elements.
First extra element 0:
<type 'object'>

- ()
+ (<type 'object'>,)

======================================================================
FAIL: Test if all the TVTK classes can be instantiated
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

Issue #233, opened 2015-07-25
Fix poly data reader intermittent failing test


dmsurti:
Refer PR#220 and #231.

The poly data reader test saving and restoring visualization test fails, but only intermittently, mostly on Windows, reported by Appveyor. 

The error trace:
```
======================================================================
FAIL: Test if saving a visualization and restoring it works.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\projects\mayavi\mayavi\tests\test_poly_data_reader.py", line 182, in test_save_and_restore
    self.check_saving(self.e, self.scene, self.bounds)
  File "C:\projects\mayavi\mayavi\tests\data_reader_test_base.py", line 75, in check_saving
    self.check(scene,bounds,error)
  File "C:\projects\mayavi\mayavi\tests\data_reader_test_base.py", line 59, in check
    self.assertEqual(numpy.allclose(got, bounds, atol=error), True, msg)
AssertionError: Got: (0.0, 1.0, 0.0, 1.0, 0.0, 2.8244394598286436e+29); expected: (0.0, 1.0, 0.0, 1.0, 0.0, 1.6)
-------------------- >> begin captured logging << --------------------
mayavi.core.registry: DEBUG: Engine [<mayavi.core.null_engine.NullEngine object at 0x00000000190FB830>] named NullEngine1 registered
--------------------- >> end captured logging << ---------------------
```

Issue #204, opened 2015-06-22
Refactor TVTK Parser


dmsurti:
Currently the TVTK parser code handles all the VTK versions we support:

* 5.10
* 6.x

The code needs to be refactored into separate modules for each VTK version. This will affect the following modules:

* `code_gen`
* `wrapper_gen`
* `vtk_module`

itziakos:
I agree, I might give it a go and make a proposal. I also think that we need to increase test coverage before we move forward with such refactoring.

PR #299, opened 2016-02-28
Add vtk 7.0 support

dmsurti:
This implements #288 to add support for VTK 7.0.

- [X] Fix custom module support
- [ ]  ~~Fix `vtkSpanSpace.Resolution` default value. Should be `100`.~~ (Haven't seen this again)


tvtk/tests/test_class_tree.py
```diff
@@ -65,12 +65,20 @@ def test_basic_vtk(self):
...
+                            'vtkTypedArray', 'vtkVariantStrictWeakOrderKey',
+                            'vtkVector', 'vtkVector2', 'vtkVector3']
+                else:
+                    expect = ['object', 'vtkColor3', 'vtkColor4', 'vtkDenseArray',
+                            'vtkQuaternion', 'vtkRect',
+                            'vtkSparseArray', 'vtkTuple',
+                            'vtkTypedArray', 'vtkVector', 'vtkVector2',
+                            'vtkVector3']
             else:
+                print( 'here in vtk 7 final else')
```
jenshnielsen: Looks like a leftover debug statement?

jenshnielsen:
:+1: wouldn't it make sense to replace `vtk_major_version == 7:` with `vtk_major_version >= 7:` to future proof for VTK8 ?

dmsurti:
@jenshnielsen Thanks for your comments/review on this PR.

> wouldn't it make sense to replace vtk_major_version == 7: with vtk_major_version >= 7: to future proof for VTK8

Every new VTK release always needs testing for compatibility, based on which we update the version checks. So I think, future proofing won't add much value.

dmsurti:
The test failure is due to #233.

jenshnielsen:
No perhaps not but it seems very unlikely that VTK8 would revert to pre VTK6 definitions which is implicit what you get as this is. In any case all you changes here apart from the ones in `test_class_tree` are basically changes to make sure that VTK7 follow the VTK6 behaviour and not the older one. These would have been unnecessary if the VTK6 (6.2) changes where done with >= originally.

dmsurti:
@jenshnielsen The vtk version handling when parsing will be handled when #204 is implemented. I agree that right now it is not the simplest implementation it could be. Thanks.

dmsurti:
appveyor fails due to #233.

integrationtests/mayavi/common.py
```diff
@@ -341,6 +341,7 @@ def compare_image_with_saved_image(src_img, img_fname, threshold=10,
                 if alt_err < min_err:
                     # image is a better match.
                     err_index = count
+                    test_failed = 0
```
prabhuramachandran: I think this is not correct, this still is a failed image as the error is > threshold.  Instead we should checkin a new image for VTK 7.

prabhuramachandran:
We should eventually fix #311 to avoid having to keep doing this.

PR #303, opened 2016-02-29
Fix get_engine options agreement

kitchoi:
- If someone switches `offscreen` on, gets an OffScreenEngine created and registered, then switches `offscreen` off again, `get_engine` would still return `OffScreenEngine` if `backend=="auto".  This doesn't make sense (or should it?).  So here I implemented a fix for this special case.

- Test for `get_engine` is made more robust by using `assertIs` instead of `assertIsInstance` as NullEngine, EnvisageEngine, OffScreenEngine are all subclass of Engine (so only L40 of the master `test_engine_manager.py` is dubious).  But this is a crucial distinction for the new test added in this PR

### DIFF
--- a/mayavi/tests/test_engine_manager.py
+++ b/mayavi/tests/test_engine_manager.py
@@ -32,37 +32,49 @@ class TestEngineManager(unittest.TestCase):
     @patch_backend("envisage")
     @patch_registry_engines({"EnvisageEngine1": EnvisageEngine()})
     def test_get_engine_backend_envisage(self):
-        self.assertIsInstance(get_engine(), EnvisageEngine)
+        self.assertIs(type(get_engine()), EnvisageEngine)
 
     @patch_backend("simple")
     @patch_registry_engines({"Engine1": NullEngine()})
     def test_get_engine_backend_simple_with_existing_engine(self):
-        self.assertIsInstance(get_engine(), Engine)
+        self.assertIs(type(get_engine()), Engine)
 
     @patch_backend("auto")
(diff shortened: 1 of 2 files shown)


prabhuramachandran:
I am a bit unsure about this.  The reason is that "auto" is supposed to mean: pick a reasonable default.  The trouble is that "reasonable" is quite ambiguous.  Maybe the solution would be to add more unambiguous backends -- say an "offscreen" backend or a "ui" backend?  Are you thinking that "auto" is just another name for "ui"?  Thoughts?

kitchoi:
Indeed "reasonable" is not well defined.  So it would be helpful to check some case studies.  For example:
```
from mayavi import mlab
mlab.options.offscreen = True
mlab.test_contour3d()
mlab.savefig("saved.png")

mlab.options.offscreen = False
mlab.figure()  ### Here
```
should the reasonable engine on the last line be an Engine (i.e. on screen) and not an OffScreenEngine?

prabhuramachandran:
Hmm, I think if offscreen is false, it should pop-up a view but then again, since there is already an off screen engine ready, technically it could use the same off screen engine.  

OK, so if offscreen is not set, one should not get an offscreen engine.  So in your case, one should get a non-offscreen engine depending on the backend.  Is this what happens with your PR?

kitchoi:
> OK, so if offscreen is not set, one should not get an offscreen engine.  So in your case, one should get a non-offscreen engine depending on the backend. Is this what happens with your PR?
Indeed this PR is aiming at this.

PR #304, opened 2016-03-01
SetSize needs to be called as an unbound method (ref: Kitware/VTK)

### LINKED ISSUE #302: Segmentation fault with mlab.savefig multiple times using off_screen_rendering and resizing image
I have tested VTK 5.8.0, 5.10.1, 6.2.0, 6.3.0 and only got seg faults for VTK 6.3.0
```
from __future__ import print_function
from mayavi import mlab


def mlab_save_figs(off_screen_rendering, **kwargs):
    for ifig in xrange(5):
        figure = mlab.figure()
        figure.scene.off_screen_rendering = off_screen_rendering

        mlab.test_contour3d()
        mlab.savefig("test{}.png".format(ifig), **kwargs)

        print("figure {} created".format(ifig))
        mlab.close()


## This is fine but the saved image dimension is different from the figure (window) size
# mlab_save_figs(True)

## This seg fault while making the second/third/... figure
# mlab_save_figs(True,  size=(100, 200))

## This won't segfault but 3/4 of the image is black
## as figure size < image size
# mlab_save_figs(False, size=(800, 700), magnification=1.0)

## This won't segfault and the image seems fine but the size is not exactly (800, 700)
# mlab_save_figs(False, size=(800, 700))

```

OS: Ubuntu 12.02 LTS 64bit
Branch: master

kitchoi:
Fix #302 
The solution is found here: https://github.com/Kitware/VTK/blob/v6.3.0/Wrapping/Python/vtk/qt4/QVTKRenderWindowInteractor.py

Should we refactor the changes we make as an extension such that we can import QVTKRenderWindowInteractor from vtk directly (and then subclass etc)?

### DIFF
--- a/integrationtests/mayavi/test_mlab_savefig.py
+++ b/integrationtests/mayavi/test_mlab_savefig.py
@@ -4,7 +4,6 @@ import unittest
 import tempfile
 
 import numpy
-from PIL import Image
 
 from mayavi import mlab
 from mayavi.core.engine import Engine
--- a/integrationtests/mayavi/test_mlab_savefig.py
+++ b/integrationtests/mayavi/test_mlab_savefig.py
@@ -1,0 +1,96 @@
+import os
+import shutil
+import unittest
+import tempfile
+
+import numpy
+from PIL import Image
+
+from mayavi import mlab
+from mayavi.core.engine import Engine
+from mayavi.core.off_screen_engine import OffScreenEngine
+from mayavi.tools.figure import savefig
+
+from common import TestCase
+
+
+def create_quiver3d():
+    x, y, z = numpy.mgrid[1:10, 1:10, 1:10]
+    u, v, w = numpy.mgrid[1:10, 1:10, 1:10]
+    s = numpy.sqrt(u**2 + v**2)
+    mlab.quiver3d(x, y, z, u, v, w, scalars=s)
+
+
+# Note: the figure(window) size is delibrately set to be smaller than
+# the required size in `savefig`, this forces the re-rendering to
+# occur and catch any potential ill rendering
+
+class TestMlabSavefigUnitTest(unittest.TestCase):
+
+    def setUp(self):
+        # Make a temporary directory for saved figures
+        self.temp_dir = tempfile.mkdtemp()
+        self.filename = os.path.join(self.temp_dir, "saved_figure.png")
+
+        # this ensures that the temporary directory is removed
+        self.addCleanup(self.remove_tempdir)
+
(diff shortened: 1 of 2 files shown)


prabhuramachandran:
This is very strange and I am not sure why this is needed.  While updating our version of the QtRWI, I noticed this but didn't realize it made such a big difference.    I will have to take a look at the sources to see why this makes a difference at all.

I think it might be good to refactor the RWI.  I thought of doing this when making the changes in #250 but it seemed to require a fair bit of work.  One way to do this would be to coordinate and refactor the upstream QtRWI. I am happy to send a PR upstream but don't have the time to do the actual refactoring now.  The main problem was that some of the upstream changes did not work well for me at all.

 The wxRWI was a lot simpler and I have removed our out-of-date version and simply import the upstream version.

kitchoi:
May I attempt the refactoring?  While you say upstream, do you mean the repository under Kitware/vtk?

prabhuramachandran:
The upstream links do not clearly document why one has to use the unbound method?  For example this link in the commit that fixes it 

http://www.vtk.org/pipermail/vtk-developers/2011-August/010284.html

and it seems unrelated to the change.

prabhuramachandran:
@kitchoi -- Sure you may certainly try to do it if you have the time (and patience) for it!  Yes I mean Kitware's gitlab repo: https://gitlab.kitware.com/vtk/vtk

prabhuramachandran:
Here is a relevant thread: http://www.vtk.org/pipermail/vtk-developers/2011-August/025776.html

prabhuramachandran:
Could you please add a small test case for this?  It would also be good to reference this thread.  I still don't quite understand why this fails with VTK 6.3.0.  Thanks!

kitchoi:
@prabhuramachandran  I have added a test case for this issue.  The test case in this PR will have conflicts with #316 and #317 should they be merged.  This is because they are all related to `mlab.savefig`.

integrationtests/mayavi/test_mlab_savefig.py
```diff
@@ -0,0 +1,95 @@
+import os
+import shutil
+import unittest
+import tempfile
+
+import numpy
+from PIL import Image
```
prabhuramachandran: This doesn't seem necessary?

kitchoi: Yes you're right.  This was copied from related PRs; I just removed the import here.

prabhuramachandran:
Thanks!  Merging.

PR #306, opened 2016-03-02
Fix mlab.move error

### LINKED ISSUE #292: mlab.move(2.0,1.0,2.0) gives error
Hi,
I'm using mayavi and I would like to move the position of my camera (translate it) withoutchanging focal point and orientation. I thought that mlab.move could do the job. mlab.move() works on my computer, but move(3,-1,-1.2) gives the following error:
incompatible dimensions for cross product (dimension must be 2 or 3)
whereas mlab.move(right = [3,0] ) gives the following error:
local variable 'xhat' referenced before assignment.

I used camera.position instead but I would be happier if I could do it directly with mlab.move.
I'm using windows and canopy. python 2.7.3
Thanks
David

kitchoi:
Fix #292 
`camera.view_plane_normal` is a tuple and so it needs to be converted to an array before being multiplied to -1

@dmsurti, @prabhuramachandran where should I put the tests for this?  NullEngine can't be used because a scene is required.  `integrationtests/mayavi/test_mlab_*`  or create a `test_camera.py` under `mayavi/tests` ?

### DIFF
--- a/integrationtests/mayavi/test_tools_camera.py
+++ b/integrationtests/mayavi/test_tools_camera.py
@@ -79,10 +79,8 @@ class TestCameraUnitTest(unittest.TestCase):
     def test_camera_move_with_right(self):
         camera = self.engine.current_scene.scene.camera
 
-        with check_attrs_change(self, camera,
-                                ("focal_point", "position")), \
-                check_attrs_do_not_change(self, camera,
-                                          ("clipping_range",)):
+        with check_attrs_change(self, camera, ("focal_point", "position")), \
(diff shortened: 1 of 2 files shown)


prabhuramachandran:
I think an integration test would work best as you need a UI.  Thanks.

kitchoi:
@prabhuramachandran  I have added tests to `integratedtests`
The failed test on AppVeyor does not seem to be related to this PR and has occurred intermittently in the past.

prabhuramachandran:
Thanks LGTM.

PR #309, opened 2016-03-02
Use newer pipeline where possible

prabhuramachandran:
(Giant?) Baby step towards using the new pipeline. This means removing any explicit calls to `vtk_obj.output` (where possible) and instead setting up pipeline connections.  This still needs cleanup but seems to be the bare minimum changes needed for all the tests to pass.  There are some places where the output is explicitly used to get information about the datasets but these could eventually be modified/refactored to use the `vtkInformation` classes.

### DIFF
--- a/integrationtests/mayavi/common.py
+++ b/integrationtests/mayavi/common.py
@@ -615,6 +615,12 @@ class TestCase(Mayavi):
     ######################################################################
     # `TestCase` interface.
     ######################################################################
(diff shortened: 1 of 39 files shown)


kitchoi:
May I ask which VTK version is this for?

dmsurti:
This is for VTK versions 6.x and later.

dmsurti:
Except issue #312, LGTM. Merging.

PR #315, opened 2016-03-16
Apply has_attribute to the output in VTKFileReader

stefanoborini:
Fixes failing tests in simphony-mayavi due to incorrect application of assign_attribute
to a PolyData dataset.

### DIFF
--- a/mayavi/sources/vtk_file_reader.py
+++ b/mayavi/sources/vtk_file_reader.py
@@ -15,6 +15,7 @@ from tvtk.api import tvtk
 # Local imports.
 from mayavi.core.pipeline_info import (PipelineInfo,
         get_tvtk_dataset_name)
+from .utils import has_attributes
 from .vtk_xml_file_reader import VTKXMLFileReader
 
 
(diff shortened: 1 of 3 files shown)


mayavi/sources/vtk_file_reader.py
```diff
@@ -75,18 +76,20 @@ def _file_path_changed(self, fpath):
...
-            # FIXME: Only the first output goes through the assign
-            # attribute filter.
-            aa = self._assign_attribute
-            self.configure_input_data(aa, outputs[0])
-            self.update_data()
-            aa.update()
-            outputs[0] = aa.output
+            # FIXME: currently handling only one output (the first one) with
+            # assign attributes.
+            for index, output in enumerate(outputs):
```
prabhuramachandran: I am not sure this will do as you expect, does your test case really have multiple outputs?  Multiple outputs will not work in Mayavi as most of our code implicitly assumes one output.  This should be revisited perhaps in the future.

stefanoborini: It only has one, but then probably it's just missing a break, or I should simply remove the for loop altogether. The critical point is the has_attributes part. If you say it's a requirement that we only have one output, then it's certainly the case. I revise it.

prabhuramachandran:
LGTM, merging.

PR #316, opened 2016-03-17
Fix savefig size

kitchoi:
Two problems that lead to images saved by `mlab.savefig` to have a different size than the one requested.  (Please see the test cases in this PR and run them against the master branch)

(1) When `wx` is used as backend and when `offscreen` is set to `True`, the `pyface.ui.wx.scene.OnSize`  is not called because the window is not lifted.

(2) When the `size` in `mlab.savefig` is given but the `magnification` is "auto", the `magnification` is changed and the `size` is also modified to match that.  However, since `magnification` is required by vtk to be an integer, rounding error may cause the `size` to be different from the one requested.
e.g. `mlab.figure(size=(100, 200)); mlab.savefig("tmp.png", size=(201, 301))` gives an image of size (201, 300) instead.

### DIFF
--- a/integrationtests/mayavi/test_mlab_savefig.py
+++ b/integrationtests/mayavi/test_mlab_savefig.py
@@ -7,6 +7,8 @@ import numpy
 from PIL import Image
 
 from mayavi import mlab
+from mayavi.core.engine import Engine
+from mayavi.core.off_screen_engine import OffScreenEngine
 from mayavi.tools.figure import savefig
 
 from common import TestCase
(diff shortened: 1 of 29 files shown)


integrationtests/mayavi/test_mlab_savefig.py
```diff
@@ -0,0 +1,179 @@
...
+        # save the figure
+        savefig(self.filename, size=(131, 217), magnification=2,
+                figure=self.figure)
+
+        # check if the image size is twice as big
+        self.check_image((434, 262))
+
+    @unittest.skipIf(os.environ.get("TRAVIS", False),
+                     ("Offscreen rendering is not tested on Travis "
+                      "due to lack of GLX support"))
```
kitchoi: I have not succeeded at testing offscreen rendering on Travis.
The tests that use offscreen rendering yield the following on Travis.
```
Xlib:  extension "RANDR" missing on display ":99.0".
The program 'python' received an X Window System error.
This probably reflects a bug in the program.
The error was 'GLXBadContextTag'.
  (Details: serial 71 error_code 157 request_code 147 minor_code 5)
  (Note to programmers: normally, X errors are reported asynchronously;
   that is, you will receive the error a while after causing it.
   To debug your program, run it with the --sync command line
   option to change this behavior. You can then get a meaningful
   backtrace from your debugger if you break on the gdk_x_error() function.)
```

kitchoi:
This should fix this bug on simphony-mayavi: https://github.com/simphony/simphony-mayavi/issues/52

itziakos:
@kitchoi can  you state the OS you did these checks?

kitchoi:
Ah right!  I tested on Ubuntu 12 64bit (via a VM), wx 2.8(.11?  not on the machine at the moment to confirm)

kitchoi:
Update:
To confirm that the version of wxPython tested is 2.8.10.1, on Ubuntu 12 LTS 64bit

Update:
This branch only deals with the size of the saved figure.   For saved figures being wrong (have black spots), see #317 which is branched from this branch.

Update:
Tests fail on Mac OS (with wxPython 2.8.10) with the master branch and with this branch (i.e. same problem exists on Mac OS and this branch does not fix it)

kitchoi:
In addition to checking out the branch and run the test cases, one can try the following scripts

```
from mayavi import mlab
import numpy

size = (140, 200)

def test(filename, off_screen_rendering, magnification=None):
    fig = mlab.figure(size=(90, 132))
    fig.scene.off_screen_rendering = off_screen_rendering
    mlab.test_contour3d()
    mlab.savefig(filename, figure=fig, size=size, magnification=magnification)


# off-screen + auto magnification
print "offsceen + auto mag"
test("bad_off_mag_auto.png", True, "auto")

# off-screen + magnification
print "offsceen + mag=1"
test("bad_off_mag_1.png", True, 1)

# off-screen + magnification = 2
print "offsceen + mag=2"
test("bad_off_mag_2.png", True, 2)

# no off-screen + auto magnification
print "no offsceen and auto mag"
test("bad_no_off_mag_auto.png", False, "auto")

# no off-screen + magnification
print "no offsceen + mag=1"
test("bad_no_off_mag_1.png", False, 1)

# no off-screen + magnification = 2
print "no offscreen + mag=2"
test("bad_no_off_mag_2.png", False, 2)
```

tvtk/pyface/ui/wx/scene.py
```diff
@@ -658,21 +658,23 @@ def _show_parent_hack(window, parent):
...
             w.SetSize((sz[0]-1, sz[1]-1))
             w.SetSize(sz)
-            window._idle_count -= 1
-            if window._idle_count < 1:
-                wx.EVT_IDLE(window, None)
-                del window._idle_count
+            # Remove the handler
+            wx.EVT_IDLE(window, None)
 
-        window._idle_count = 2
```
kitchoi: This change regarding the idle event handler is for fixing the saved image size on Mac OS with wxPython 2.9.2.4 and 3.0.2.0 (tested on OS X 10.11.2).

I don't know why it was better to call the resize function for the first *two* Idle events and not just the first one idle event.  Having tested on Ubuntu and MacOS with wxPython 2.8+, I have not encountered a situation when doing the resize once is insufficient.  Was this for some older wx versions?

Nevertheless, given that `mlab.savefig(size=size)` is the next thing to do after creating a scene, doing the resize for the second idle event results in an image of the wrong size on Mac OS.  This is caused by the idle event handler being called after `scene.set_size` and before the `scene.save_png` (or save_bmp... etc) function inside `scene.save`; i.e. on Mac OS, the event handlers are fast and efficient - nothing wrong on wx part.  For some reason the event handling is slower on Ubuntu (e.g. Travis) and so the size of the saved image was fine.   My OS X version is also recent and the Ubuntu is running on a VM with fewer CPU cores - this could be why.

kitchoi:
This PR regarding the image size of `mlab.savefig` is ready for review.

tvtk/pyface/ui/wx/scene.py
```diff
@@ -681,9 +683,12 @@ def _do_idle(event, window=window):
...
         image."""
-        if self.render_window.off_screen_rendering:
-            # Do nothing if off screen rendering is being used.
-            return
+        ## Even when off_screen_rendering is On, the image size would
+        ## be wrong if we don't lift the window because the OnSize
+        ## event would not be called anymore
+        # if self.render_window.off_screen_rendering:
+        #     # Do nothing if off screen rendering is being used.
+        #     return
```
kitchoi: This changes fixes the size on Ubuntu but leads to wrong images on Mac OS.
I realise this PR #316 is now interwined with #317.

kitchoi: I still think the render window should be lifted for the size to be correct.   The reason why this would make saved images to have black spots on Mac OS is precisely due to resizing and should be tackled.

kitchoi:
Superseded by #331, closing.

Issue #75, opened 2013-06-08
offscreen rendering with Xvfb:  strange results, only black images. (Fedora 18, mayavi 4.3 installed with rpm, or installed from the source on github), 


goballooning:
(Sorry that I'm not quite sure whether it is the right place to report this issue.)

I render some images using `Xvfb` according to the tips here http://docs.enthought.com/mayavi/mayavi/tips.html#offscreen-rendering. 

It worked perfectly under Fedora 17, mayavi 4.2.
But on Fedora 18, with all package updated, and 
- Mayavi.x86_64 0:4.3.0-2.fc18., installed with yum, 
- Mayavi 4.3, installed from source on github. 

for these two cases, the saved images are always black.

After removing mayavi 4.3, and installing mayavi 4.2 from rpmfind.net, there is no such black image problem any more.

Hongying

stefanoborini:
I experienced the same problem on ubuntu 14 and mayavi 4.4.4

itziakos:
I am picking this up

kitchoi:
I suspect this has to do with the `read_front_buffer` flag in `tvtk.WindowToImageFilter` in `tvtk.pyface.tvtk_scene`.
On travis-python (docker container) + xvfb + offscreen-off, this issue is reproduced and is resolved by setting `read_front_buffer=False`.

Here are the commits for reference where this flag is modified:
Mayavi 4.2 to 4.3: https://github.com/enthought/mayavi/commit/ed295f799e6b7edba9ba6cfc3858058cca99883a
And here: https://github.com/enthought/mayavi/pull/76/commits/9588883aed301a20a8ff3b618fecca8963a545d5

PR #317, opened 2016-03-17
Fix erroneous images when snapshot size is larger than window

kitchoi:
Added tests looking out for black spots in an image that should not have any.
In addition to running the added test cases against the current master branch, one can see the problem on their own machine by doing this:
```
from mayavi import mlab

fig = mlab.figure(size=(90, 130))
mlab.test_contour3d()
# notice that the image size is larger than the window size
mlab.savefig("saved_file.png", size=(140, 200), magnification=2)
```
I have tested this branch with a headed machine under xvfb.

Potentially related to #75, effectively undoing #58 

OS: Ubuntu 12.04 LTS 64bit
VTK: 6.2.0
wxPython: 2.8.10

### DIFF
--- a/integrationtests/mayavi/test_mlab_savefig.py
+++ b/integrationtests/mayavi/test_mlab_savefig.py
@@ -162,6 +162,11 @@ class TestMlabSavefigUnitTest(unittest.TestCase):
         if size:
             self.assertEqual(image.shape[:2], size)
 
+        # check if the image has black spots
+        if (numpy.sum(image == [0, 0, 0], axis=2) == 3).any():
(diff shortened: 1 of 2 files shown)


kitchoi:
Update: this fix does not work on Mac OS X 10.11 with wxPython 2.9.2.4 and 3.0.2.0  Need more investigation.

kitchoi:
This fixes black images on Linux machine.  Black images on OS X appears to be a separate issue.  Will make a separate issue for it.

kitchoi:
Closing.  Opening a new PR that supersedes this.

PR #323, opened 2016-03-23
Enforce ordering of attribute labels in VTKDataSource.

stefanoborini:
Currently, the entries have no guarantee of order, as they come from a dictionary.
We enforce the sorted order for user friendliness and consistency.


prabhuramachandran:
LGTM.

PR #324, opened 2016-03-23
Refactor QVTKRenderWindowInteractor [WIP]

kitchoi:
Attempt to use the vtk.qt4.QVTKRenderWindowInteractor when possible.  It also makes clear the patches we needed.  Some of these patches may be sent upstream to VTK.  However, they are likely to be kept around for sometime unless there are plans for discontinuing supports for older VTK versions.

This would supersede #304 (except for the test case added there).

### DIFF
--- a/tvtk/pyface/ui/qt4/QVTKRenderWindowInteractor.py
+++ b/tvtk/pyface/ui/qt4/QVTKRenderWindowInteractor.py
@@ -174,6 +174,14 @@ class QVTKRenderWindowInteractor(_QVTKRenderWindowInteractor):
             WId = pythonapi.PyCapsule_GetPointer(WId, name)
         return str(int(WId))
 
+    def Finalize(self):
+        '''
+        Call internal cleanup method on VTK objects
+
+        New in VTK 6.0.0
+        '''
+        self._RenderWindow.Finalize()
+
     def paintEvent(self, ev):
         super(QVTKRenderWindowInteractor, self).paintEvent(ev)
         self._RenderWindow.Render()
--- a/tvtk/pyface/ui/qt4/QVTKRenderWindowInteractor.py
+++ b/tvtk/pyface/ui/qt4/QVTKRenderWindowInteractor.py
@@ -3,7 +3,6 @@ A simple VTK widget for PyQt or PySide.
 See http://www.trolltech.com for Qt documentation,
 http://www.riverbankcomputing.co.uk for PyQt, and
 http://pyside.github.io for PySide.
-
 This class is based on the vtkGenericRenderWindowInteractor and is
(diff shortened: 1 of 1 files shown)


tvtk/pyface/ui/qt4/QVTKRenderWindowInteractor.py
```diff
@@ -271,50 +177,17 @@ def _get_win_id(self):
...
-
-    def sizeHint(self):
-        return QSize(400, 400)
-
-    def paintEngine(self):
-        return None
-
     def paintEvent(self, ev):
+        super(QVTKRenderWindowInteractor, self).paintEvent(ev)
         self._RenderWindow.Render()
```
kitchoi: I am not sure if this `self._RenderWindow.Render()` is only required by tvtk Scene.  If so, then it should be moved to tvtk/pyface/ui/qt4/scene.py

tvtk/pyface/ui/qt4/QVTKRenderWindowInteractor.py
```diff
@@ -271,50 +177,17 @@ def _get_win_id(self):
     def Finalize(self):
```
prabhuramachandran: IIRC the finalize used to cause a crash on OS X, I can't remember exactly how right now but can you test this version please to ensure that when you create a window and close it, it doesn't crash.  Thanks!

kitchoi: I tried
```
from mayavi import mlab
mlab.figure()
mlab.close()
```
with this branch on OS X 10.11.2, PySide/PyQt with Qt 4.8.7 and there is no crash.
Also run the integratedtests and no crash either.  Note that my OS X version is fairly up-to-date...

tvtk/pyface/ui/qt4/QVTKRenderWindowInteractor.py
```diff
@@ -174,19 +89,17 @@ def __init__(self, parent=None, wflags=Qt.WindowFlags(), **kw):
             rw = None
 
         # create qt-level widget
-        QWidget.__init__(self, parent, wflags|Qt.MSWindowsOwnDC)
+        QWidget.__init__(self, parent, wflags | Qt.MSWindowsOwnDC)
```
prabhuramachandran: Maybe a comment as to why we are not calling the parent's __init__ would be good.

kitchoi:
I just tested the branch with integrationtests, and it no longer works.  :(  I will have to revisit this.

prabhuramachandran:
@kitchoi -- is it OK to close this now?  The interactor code has also changed a bit now and more inline with the version in VTK.  Please reopen if you need to and the issues are resolved.  Thanks!

PR #329, opened 2016-03-30
Suppresses test for VTK 6.2 (solves appVeyor random failure)

stefanoborini:
We have occasional appveyor failure which is probably due to the PLY format
still not working.


stefanoborini:
This should also solve current failing PRs. Please merge this one first, and will be brought into these PRs.

kitchoi:
Perhaps add an issue number to the skip message?

stefanoborini:
@kitchoi done. Merge?

kitchoi:
I don't think I know enough about the test case (its objective, why it only fails on VTK 5.6/5.8 and Windows) to say.  @dmsurti, @prabhuramachandran?

prabhuramachandran:
Fixing this will require some careful investigation.  I guess this is fine for now as this particular error is getting to be irritating.

PR #331, opened 2016-03-31
Fix savefig size and black image

### LINKED ISSUE #257: offscreen rendering outputs black image unless new renderwindow created
This occurs on OSX 10.9.5 (AMD 6750M graphics card) with mayavi 4.4.2.

A simple test of off screen rendering and image saving, such as:

    from mayavi import mlab
    
    mlab.options.offscreen = True
    mlab.test_contour3d()
    fig = mlab.gcf()
    mlab.savefig(figure=fig, filename='example.png')

gives a blank output (i.e. a black image).

However, if a new render window is created and then the figure's renderer is added to this render window, for example (based on tvtk offscreen example code, could undoubtedly be cleaned up):

    from mayavi import mlab
    from tvtk.api import tvtk
    from tvtk.common import configure_input_data
    
    mlab.options.offscreen = True
    mlab.test_contour3d()
    fig = mlab.gcf()
    rw = tvtk.RenderWindow(size=fig.scene._renwin.size, off_screen_rendering=1)
    rw.add_renderer(fig.scene._renderer)

    w2if = tvtk.WindowToImageFilter()
    w2if.magnification = fig.scene.magnification
    w2if.input = rw
    ex = tvtk.PNGWriter()
    ex.file_name = 'example4.png'
    configure_input_data(ex, w2if.output)
    w2if.update()
    ex.write()

the image output is correct. 

From a brief inspection it's not clear as to how the new render window differs from the one that is automatically generated.

Regardless of whether the image is saved correctly, the rendering seems to be much slower when `mlab.options.offscreen = True` than when `mlab.options.offscreen = False`.

kitchoi:
Fix https://github.com/simphony/simphony-mayavi/issues/52, fix #257, potentially related to #75

Supersede #316 and #317 (the two issues are highly related that it has become impossible to separate the two)

Upon saving an image (snapshot), a new tvtkRenderWindow is created.  This is motivated by the fact that under Mac OS, vtkCocoaRenderWindow is the render window being used and if the vtkRenderWindow is contained by an interactor, resizing the renderwindow leads to the resulting image to be rendered only within the initial size.  Therefore a new tvtkRenderWindow is needed for saving an image in Mac.

This fix for Mac causes a side effect for Linux which is that when offscreen is off, saving an image will create a new window.  Forcing offscreen to true for the dummy tvtkRenderWindow would enhance the user experience but render the savefig function completely unusable if the machine does not support offscreen.

Tested platforms:
Mac OS 10.11 + wxPython/PyQt + VTK 6.2
Ubuntu 12.02 LTS + VTK 5.10/6.2.0/6.3.0 + wxPython/PyQt

### DIFF
--- a/tvtk/pyface/ui/wx/scene.py
+++ b/tvtk/pyface/ui/wx/scene.py
@@ -687,12 +687,9 @@ class Scene(TVTKScene, Widget):
     def _lift(self):
         """Lift the window to the top. Useful when saving screen to an
         image."""
-        ## Even when off_screen_rendering is On, the image size would
-        ## be wrong if we don't lift the window because the OnSize
-        ## event would not be called anymore
-        # if self.render_window.off_screen_rendering:
-        #     # Do nothing if off screen rendering is being used.
-        #     return
+        if self.render_window.off_screen_rendering:
+            # Do nothing if off screen rendering is being used.
+            return
 
         w = self._vtk_control
         while w and not w.IsTopLevel():
--- a/tvtk/pyface/ui/wx/scene.py
+++ b/tvtk/pyface/ui/wx/scene.py
@@ -571,6 +571,10 @@ class Scene(TVTKScene, Widget):
     def _closed_fired(self):
         super(Scene, self)._closed_fired()
         self.picker = None
+        # Remove OnPaint handler for PaintEvent, otherwise
+        # OnPaint tries to reset the size of the nonexisting
+        # renderwindow
+        wx.EVT_PAINT(self._vtk_control, None)
         self._vtk_control = None
 
     ###########################################################################
--- a/tvtk/pyface/ui/wx/scene.py
+++ b/tvtk/pyface/ui/wx/scene.py
@@ -329,6 +329,9 @@ class Scene(TVTKScene, Widget):
 
     def set_size(self, size):
         """Set the size of the window."""
+        # the OpenGLRenderWindow needs resizing too; otherwise snapshots
+        # with off_screen_rendering would have the wrong size
+        self._vtk_control._Iren.GetRenderWindow().SetSize(size)
         self._vtk_control.SetSize(size)
 
     def hide_cursor(self):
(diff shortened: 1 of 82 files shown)


kitchoi:
There may be a better way.

kitchoi:
The "better" way doesn't work.  Never mind.

kitchoi:
This is good for review.

stefanoborini:
Reviewing this one

tvtk/pyface/ui/wx/scene.py
```diff
@@ -571,6 +571,10 @@ def OnButtonUp(self, event):
     def _closed_fired(self):
         super(Scene, self)._closed_fired()
         self.picker = None
+        # Remove OnPaint handler for PaintEvent, otherwise
+        # OnPaint tries to reset the size of the nonexisting
+        # renderwindow
+        wx.EVT_PAINT(self._vtk_control, None)
```
stefanoborini: This one makes me a bit nervous because I checked the rest of the class, and nowhere it is enforced that self._vtk_control is not None. We have the same problem pretty much everywhere. I am generally fine with throwing an exception, but if None is an allowed value, we should react appropriately without throwing. If it's not, it's better to enforce the presence of a control at construction time.

kitchoi: `_create_control` is called here https://github.com/enthought/mayavi/blob/fix-savefig-size-black-image/tvtk/pyface/tvtk_scene.py#L205  which assigns `_vtk_control`.  Unless `self._vtk_control = None` is done somewhere outside of this function (which I have not checked), we can ensure `_vtk_control` is not None?

kitchoi: On other hand, this change is meant to silent error message on Linux + wx 2.8.10
https://github.com/enthought/mayavi/pull/346
The same error does not occur on OSX with wxpython 3.0.2, so I have omitted this change in the reimplementation #346

stefanoborini: Good. Merging.

kitchoi:
This fix requires more investigation as the implementation is not ideal.
See #346, #347

PR #335, opened 2016-04-05
Fix texture

### LINKED ISSUE #211: texture on mlab.surf
I am trying to place a texture on a surface (mlab.surf).  Instead, I see a surface with the mean pixel color of that texture image. I would appreciate a confirmation that this great feature (texture on mlab.surf) should still work as advertised and my installed system is to blame.

Similar code worked for me with mayavi 4.3 and vtk5 on a debian system. I did change the arguments passed to tvtk.Texture for it to work with mayavi 4.4.2 and vtk6 (via homebrew and virtenv+pip on OS X).

Thanks in advance!

### EXAMPLE:

```python
from tvtk.api import tvtk
from mayavi import mlab
import numpy as np

mlab.clf()

X,Y=np.mgrid[-1:1:20j,-1:1:20j]
Z = -np.cos(Y*X)+.5

img_filename='anyimage.jpg'
img=tvtk.JPEGReader(file_name=img_filename)
texture=tvtk.Texture(input_connection=img.output_port, interpolate=1)

# following produces a surface with a color that seems to be represent the mean img pixel.
s = mlab.surf(X,Y,Z,color=(1.,1.,1.))
s.actor.actor.mapper.scalar_visibility=False
s.actor.enable_texture=True
s.actor.tcoord_generator_mode='plane'
s.actor.actor.texture=texture

# but the texture itself works!
from mayavi.sources.api import BuiltinSurface
p = mlab.pipeline.surface( BuiltinSurface(source='plane') )
p.actor.actor.texture=texture```

kitchoi:
Fix #211


prabhuramachandran:
Thanks for the careful test!

lmkli:
I tried and can only get the correct texture on the ''cylinder" case, while the other two are both failed. Any idea?

PR #340, opened 2016-04-11
Import and integer division in python3

kitchoi:
Two small fixes in compliance with Python3

(1) Explicit relative import for a test
(2) Integer division may cause error, use `-1` to infer the size of the remaining dimension in numpy.reshape


prabhuramachandran:
LGTM.

PR #341, opened 2016-04-11
Fix traitsdoc setup and compliance with py3

### LINKED ISSUE #228: The `python setup.py build_docs`  command does not build on windows with sphinx 1.2.3
It looks that we need to revisit the `build_docs` command in the `setup.py`. Recent versions of sphinx support the `build_sphinx` command and one can run `python setup.py build_sphinx` but it is not clear how well it works.

kitchoi:
Traits properties should show in the API documentation with this PR.  And doc build should work on Python3.

Although previously changes in #293 led to successful doc build, it in fact did not setup sphinx to use `traitsdoc.get_doc_object` and `traitsdoc.SphinxTraitsDoc` (my fault!).  This PR fixes this problem.

Partial fix is made to `compiler_unparse` according changes in the API of `ast`; enough for the current doc build.  In the future we may consider using customised comment syntax for documenting traits (e.g. https://github.com/enthought/trait-documenter) instead of printing the code for the traits definition, therefore removing the need for unparsing AST back to code.

Tested on OS X + Python3 + Sphinx 1.4, and Ubuntu + Python2.7 + Sphinx1.3.5

@itziakos : does this fix #228 too?

### DIFF
--- a/docs/source/mayavi/sphinxext/sphinxext/comment_eater.py
+++ b/docs/source/mayavi/sphinxext/sphinxext/comment_eater.py
@@ -169,9 +169,6 @@ def get_class_traits(klass):
     # AST tree for the class
     class_ast = mod_ast.body[0]
 
-    # Source code needed for getting the right-hand side
-    source_code = source.split("\n")
-
     # All assignement nodes
     assign_nodes = [node for node in class_ast.body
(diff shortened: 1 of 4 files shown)


prabhuramachandran:
Thanks, this is awesome!  Merging.

PR #342, opened 2016-04-13
Fix: Nicely load pickled pylab_luts.pkl

kitchoi:
(Update: this PR is mostly a bug fix than an enhancement)
The current pylab_luts.pkl could not be loaded on Windows 64bit + Python 2.7 64bit (tested with both cmd.exe and PowerShell, and with virtualenv)
```
>>> fpath = "mayavi\\core\\lut\\pylab_luts.pkl"
>>> state = state_pickler.load_state(fpath)
  File "<stdin>", line 1, in <module>
  File "C:\Users\kit\envs\test-mayavi\lib\site-packages\apptools\persistence\state_pickler.py", line 959, in load_state
    state = StateUnpickler().load_state(f)
  File "C:\Users\kit\envs\test-mayavi\lib\site-packages\apptools\persistence\state_pickler.py", line 515, in load_state
    data = pickle.load(file)
  File "c:\python27\Lib\pickle.py", line 1384, in load
    return Unpickler(file).load()
  File "c:\python27\Lib\pickle.py", line 864, in load
    dispatch[key](self)
  File "c:\python27\Lib\pickle.py", line 972, in load_string
    raise ValueError, "insecure string pickle"
ValueError: insecure string pickle
```

Since `pylab_luts.pkl` is simply a dictionary with numpy arrays, I propose using the standard library `pickle` for this task.

Side issue, I could not see logger.warning message in Mayavi2 while I was testing this PR with a faulty pylab_luts.pkl.  Perhaps the logger is not set up properly or I am doing something stupid :/


kitchoi:
This PR also means to turn the Error on loading pylab_luts.pkl into a warning because failing to load pylab_luts only leads to fewer available colormap, i.e. not catastrophic enough to crash.

kitchoi:
I realise the unpickling problem I encountered is not due to the use of state_pickler but likely due to the differences in OS architecture.  state_pickler compresses the data and creates a smaller file which is desirable.  Closing this for now.  Needs more investigation.

PR #343, opened 2016-04-18
fixes for the tvtk_segmentation.py example

grlee77:
The first commit here fixes the following error:

```
  File "examples/mayavi/advanced_visualization/tvtk_segmentation.py", line 90, in <module>
    median_filter.set_kernel_size(3, 3, 3)
AttributeError: 'ImageMedian3D' object has no attribute 'set_kernel_size'
```

The second avoids the following warnings raised by numpy 1.11

```
examples/mayavi/advanced_visualization/tvtk_segmentation.py:52: VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future
  lower_thr = sorted_data[0.2*l]
examples/mayavi/advanced_visualization/tvtk_segmentation.py:53: VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future
  upper_thr = sorted_data[0.8*l]
```

### DIFF
--- a/examples/mayavi/advanced_visualization/tvtk_segmentation.py
+++ b/examples/mayavi/advanced_visualization/tvtk_segmentation.py
@@ -49,8 +49,8 @@ data = data.T
 # scipy.stats.scoreatpercentile)
 sorted_data = np.sort(data.ravel())
 l = len(sorted_data)
-lower_thr = sorted_data[0.2*l]
-upper_thr = sorted_data[0.8*l]
+lower_thr = sorted_data[int(0.2*l)]
+upper_thr = sorted_data[int(0.8*l)]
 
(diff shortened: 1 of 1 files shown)


prabhuramachandran:
@grlee77 -- Thanks for this, can you tell us what version of VTK you are using?  The change to `set_kernel_size` does not work with VTK 6.x.

grlee77:
I see.  I am on VTK 7.0

grlee77:
I am not very familiar with the mayavi codebase and conventions.  What is the preferred way to handle differences across VTK versions?

prabhuramachandran:
Thanks @grlee77 -- I have fixed this in #542 so am closing this PR, thank you!

PR #344, opened 2016-04-19
Allow load pylab_luts.pkl to fail

kitchoi:
If for some reasons, unpickling `pylab_luts.pkl` fails (had it failed on Windows 64bit), it should not be fatal: just fewer colormaps available.

Added a small test for this.


mayavi/core/lut_manager.py
```diff
@@ -22,9 +22,30 @@
...
-                                                'pylab_luts.pkl'))
+pylab_luts_file = os.path.join(lut_image_dir, 'pylab_luts.pkl')
+
+try:
+    pylab_luts = state_pickler.load_state(pylab_luts_file)
+except (IOError, ValueError) as exception:
+    # IOError: failed to open file
+    # ValueError: pickled file is built from an OS w/ different architecture,
+    #             or with an incompatible protocol
+    message = ("Failed to load pylab colormaps from file: {filepath}\n"
```
prabhuramachandran: I think we should dump os specific files and load them in that case instead of failing and asking the user to build this.

kitchoi:
Going to open a new PR without the binary file so not to pollute history.  Closing this one.

PR #345, opened 2016-04-19
Add VTK 6.3 and VTK 7 to travis ci

kitchoi:
Use VTK official build of vtkPython 6.3 and 7.0.
Currently they are allowed to fail on travis.

For VTK 7.0, get lots of X windows errors there, which might be wx problem.  I set ETS_TOOLKIT=wx for running the integrationtests.  There are some genuine errors highlighting changes to the new pipeline.


kitchoi:
The last few commits were a bit messy, I thought I could run integrationtests with pyside given Ubuntu has a `python-pyside` package; turns out it is not white-listed on travis for Ubuntu 12 (but is whitelisted for Ubuntu 14).  So I added a few comments to .travis.yml to explain why we aren't running integrationtests with qt4 on Travis CI, in case someone else wants to try this again.
@stefanoborini could you review this please?

PR #349, opened 2016-04-21
Revert "Fix savefig size and black image"

kitchoi:
Reverts enthought/mayavi#331
This should not be merged yet.

### DIFF
--- a/integrationtests/mayavi/test_mlab_savefig.py
+++ b/integrationtests/mayavi/test_mlab_savefig.py
@@ -4,7 +4,6 @@ import unittest
 import tempfile
 
 import numpy
-from PIL import Image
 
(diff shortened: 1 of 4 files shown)


stefanoborini:
oh ok... sorry, but does it work?

kitchoi:
Works except when on-screen + the size of image requested is larger than my physical screen resolution.  It wasn't captured in the test_mlab_savefig.py as the image size was not big enough, and was only recently discovered (today)

stefanoborini:
Ok. Merging when green, please restore the branch and PR for the other one. Leave it [WIP]

kitchoi:
Appveyor has hung but Travis passed.  I will proceed with the revert now (sooner the better).

PR #347, opened 2016-04-21
Reimplement fix savefig size and black image - step 1 [WIP]

kitchoi:
This is the next step after fixing the size of the snapshot images revealing black images (#346). 
This branch was implemented as #317 but was then reverted and superseded by (#331) 
Effectively undoing #58
As offscreen tests cannot be run on Travis, these tests had to be run locally with a headed machine.

OSX: 10.11, PySide 1.2.2, wxPython 3.0.2.0
Linux: Ubuntu 12, PySide 1.2.2, wxPython 2.8.10.1

**Linux + qt4 + this branch** All passed!

**Linux + wx + this branch** (On Travis these offscreen tests are skipped)
```
======================================================================
FAIL: test_savefig_with_size_and_magnification_offscreen (test_mlab_savefig.TestMlabSavefigUnitTest)
Test savefig with given size, mag, normal Engine and offscreen
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_mlab_savefig.py", line 183, in test_savefig_with_size_and_magnification_offscreen
    self.check_image_size(self.filename, size=(262, 434))
  File "test_mlab_savefig.py", line 82, in check_image_size
    self.assertEqual(image.shape[:2][::-1], size)
AssertionError: Tuples differ: (180, 140) != (262, 434)

First differing element 0:
180
262

- (180, 140)
+ (262, 434)

======================================================================
FAIL: test_savefig_with_size_offscreen (test_mlab_savefig.TestMlabSavefigUnitTest)
Test savefig with given size, normal Engine and offscreen
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_mlab_savefig.py", line 162, in test_savefig_with_size_offscreen
    self.check_image_size(self.filename, size=(131, 217))
  File "test_mlab_savefig.py", line 82, in check_image_size
    self.assertEqual(image.shape[:2][::-1], size)
AssertionError: Tuples differ: (90, 70) != (131, 217)

First differing element 0:
90
131

- (90, 70)
+ (131, 217)

----------------------------------------------------------------------
Ran 7 tests in 5.161s

FAILED (failures=2)
```

**OSX + qt4 + this branch**
```
======================================================================
FAIL: test_savefig_with_size (test_mlab_savefig.TestMlabSavefigUnitTest)
Test if savefig works with given size and a normal Engine
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_mlab_savefig.py", line 109, in test_savefig_with_size
[...]

PR #346, opened 2016-04-21
Reimplement fix savefig with size and magnification - step 0 [WIP]

kitchoi:
Reimplement part of #331 
Tests fail as the fix is incomplete.  I aim at making a new branch from this one for fixing those new problems, so that we know what was attempted and what does not work.  This issue is complicated.

Running `integratedtests/mayavi/test_mlab_savefig.py` against **master** leads to the following errors on *both Ubuntu 12 and OSX 10.11*

OSX: 10.11, PySide 1.2.2, wxPython 3.0.2.0
Linux: Ubuntu 12, PySide 1.2.2, wxPython 2.8.10.1

**Linux/OSX + qt4 + master branch**
```
======================================================================
FAIL: test_savefig_with_size (test_mlab_savefig.TestMlabSavefigUnitTest)
Test if savefig works with given size and a normal Engine
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_mlab_savefig.py", line 108, in test_savefig_with_size
    self.check_image_size(self.filename, size=(131, 217))
  File "test_mlab_savefig.py", line 82, in check_image_size
    self.assertEqual(image.shape[:2][::-1], size)
AssertionError: Tuples differ: (128, 216) != (131, 217)

First differing element 0:
128
131

- (128, 216)
+ (131, 217)

======================================================================
FAIL: test_savefig_with_size_and_magnification (test_mlab_savefig.TestMlabSavefigUnitTest)
Test if savefig works with given size and magnification
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_mlab_savefig.py", line 124, in test_savefig_with_size_and_magnification
    self.check_image_no_black_pixel(self.filename)
  File "test_mlab_savefig.py", line 72, in check_image_no_black_pixel
    self.fail(message)
AssertionError: The image has black spots

======================================================================
FAIL: test_savefig_with_size_offscreen (test_mlab_savefig.TestMlabSavefigUnitTest)
Test savefig with given size, normal Engine and offscreen
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_mlab_savefig.py", line 156, in test_savefig_with_size_offscreen
    self.check_image_size(self.filename, size=(131, 217))
  File "test_mlab_savefig.py", line 82, in check_image_size
    self.assertEqual(image.shape[:2][::-1], size)
AssertionError: Tuples differ: (128, 216) != (131, 217)

First differing element 0:
128
131

- (128, 216)
+ (131, 217)

[...]

PR #350, opened 2016-04-21
Reimplement fix savefig size and black image - step 2 (last)

kitchoi:
#346, #347 and this complete the whole journey to #331 
(I would leave #331 close and this one opens)

If this whole thing should be merged, this PR merges to the branch of #347, and then #347 will merge to the branch of #346.

### DIFF
--- a/integrationtests/mayavi/test_mlab_savefig.py
+++ b/integrationtests/mayavi/test_mlab_savefig.py
@@ -1,3 +1,4 @@
+import sys
 import os
 import shutil
 import unittest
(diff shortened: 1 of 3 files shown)


kitchoi:
Recent commits differentiate this PR from #331

kitchoi:
With this fix, upon saving an image with a specific size while the window is on-screen (i.e. not offscreen), a new window is opened and then closed.  This is not ideal and I would very much appreciate any suggestion for an alternative.  Note that the problem this PR is trying to fix affects Linux+wx and Mac OS (qt4/wx).

kitchoi:
Stale, closing.

PR #352, opened 2016-04-22
ENH: Nice load pylab_luts.pkl

kitchoi:
Supersede #342  as a new binary file is not required.

The original problem arises when git checkout the repository on Windows, *.pkl files are not thought as binary file and its line endings are converted.  Adding `.gitattributes` solves the problem.  On the other hand, I still think soft loading the pickle file is appropriate, for unforseen failures.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+*.png binary
+*.jpg binary
+*.pdf binary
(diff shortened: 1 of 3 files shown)


prabhuramachandran:
LGTM thanks!

Issue #354, opened 2016-04-22
pip install mayavi fail


kitchoi:
In a clean virtual environment (with Python 2.7, Linux 64bit), checking out the repository and do `python setup.py install` works, but not with `pip install`.

Assuming you have installed vtk as well (here I tried vtkPython 6.3.0 downloaded from vtk.org)
```
$ pip install numpy
$ pip install mayavi

...[skip many lines]...
  ----------------------------------------------------------------------
  Building TVTK classes... vtkGeoTreeNode: Ignoring method: Get/SetStatus
  default: 0, range: None

...[skip more lines]...

  Traceback (most recent call last):
    File "<string>", line 1, in <module>
    File "/tmp/pip-build-TXVQcu/mayavi/setup.py", line 476, in <module>
      **config
    File "/home/kit/.virtualenvs/test/local/lib/python2.7/site-packages/numpy/distutils/core.py", line 169, in setup
      return old_setup(**new_attr)
    File "/usr/lib/python2.7/distutils/core.py", line 152, in setup
      dist.run_commands()
    File "/usr/lib/python2.7/distutils/dist.py", line 953, in run_commands
      self.run_command(cmd)
    File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
      cmd_obj.run()
    File "/home/kit/.virtualenvs/test/local/lib/python2.7/site-packages/wheel/bdist_wheel.py", line 179, in run
      self.run_command('build')
    File "/usr/lib/python2.7/distutils/cmd.py", line 326, in run_command
      self.distribution.run_command(command)
    File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
      cmd_obj.run()
    File "/tmp/pip-build-TXVQcu/mayavi/setup.py", line 308, in run
      build_tvtk_classes_zip()
    File "/tmp/pip-build-TXVQcu/mayavi/setup.py", line 294, in build_tvtk_classes_zip
      gen_tvtk_classes_zip()
    File "tvtk/setup.py", line 66, in gen_tvtk_classes_zip
      import numpy
    File "/tmp/pip-build-TXVQcu/mayavi/tvtk/code_gen.py", line 117, in generate_code
      self._write_wrapper_class(node, tvtk_name)
    File "/tmp/pip-build-TXVQcu/mayavi/tvtk/code_gen.py", line 207, in _write_wrapper_class
      self.wrap_gen.generate_code(node, out)
[...]

PR #355, opened 2016-04-25
Unicode trait handling for those vtk methods returning unicode data

stefanoborini:
The generator code did not handle unicode, but some vtk routines do return them. Although never able to reproduce it, issue #354 shows a situation where it does matter.

I am unsure about potential side effects of this change on client code, but I am surprised it worked until now, and I expect the method was simply not generated, or it was but it returned an empty string.

### DIFF
--- a/tvtk/wrapper_gen.py
+++ b/tvtk/wrapper_gen.py
@@ -599,16 +599,16 @@ class WrapperGenerator:
                     if default == '\x00':
                         default = ''
 
-                    t_def = 'traits.String({0!r}, '.format(default)
-                    t_def += 'enter_set=True, auto_set=False)'
+                    t_def = ('traits.String({0!r}, '
+                             'enter_set=True, auto_set=False)'.format(default))
                     self._write_trait(out, name, t_def, vtk_set_meth, mapped=False)
 
                 elif PY_VER < 3 and typ is unicode:
                     if default == u'\x00':
                         default = ''
 
-                    t_def = 'traits.Unicode({0!r}, '.format(default)
-                    t_def += 'enter_set=True, auto_set=False)'
+                    t_def = ('traits.Unicode({0!r}, '
+                             'enter_set=True, auto_set=False)'.format(default))
                     self._write_trait(out, name, t_def, vtk_set_meth, mapped=False)
 
                 elif typ in (tuple,):
--- a/tvtk/wrapper_gen.py
+++ b/tvtk/wrapper_gen.py
@@ -605,7 +605,7 @@ class WrapperGenerator:
 
                 elif PY_VER < 3 and typ is unicode:
                     if default == u'\x00':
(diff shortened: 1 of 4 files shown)


tvtk/wrapper_gen.py
```diff
@@ -620,6 +620,22 @@ def _gen_get_set_methods(self, klass, out):
...
                                       mapped=False)
+                elif PY_VER < 3 and typ is unicode:
+                    default = default.encode("unicode_escape")
+                    if '\n' in default or '\r' in default:
+                        default = clean_special_chars(default)
+
+                    if default == '\x00':
+                        t_def = 'traits.Unicode(u"", '
+                    elif default == '"':
+                        t_def = "traits.Unicode(u'%(default)s', " % locals()
```
rkern:
```
if '\n' in default or '\r' in default:
    default = clean_special_chars(default)
if default == '\x00':
    default = u''
t_def = 'traits.Unicode({0!r}, enter_set=True, auto_set=False)'.format(default)
```

rkern: Don't `.encode()` or `clean_special_chars()`. `{!r}` takes care of everything.

stefanoborini: @rkern done. I also took care of the part handling String and removed the cleaning function.

kitchoi:
Verified that it works with the following tox.ini
```
[tox]
envlist = py27

[testenv]
# To make sure we actually test the tarball and now the working directory
changedir = .tox

# We use symbolic links to provide vtk which is not on pypi
whitelist_externals=/bin/ln

# Install development requirements
deps = numpy

# e.g. export VTK_PATH=${HOME}/.cache/VTK-6.3.0-Linux-64bit/lib/python2.7/site-packages/vtk
# e.g. export LD_LIBRARY_PATH=${HOME}/.cache/VTK-6.3.0-Linux-64bit/lib
passenv = VTK_PATH LD_LIBRARY_PATH

basepython = python2.7

commands =
	 ln -sf {env:VTK_PATH} {envsitepackagesdir}/
```

stefanoborini:
Failure seems due to a network problem. Restarting

tvtk/wrapper_gen.py
```diff
@@ -605,21 +596,21 @@ def _gen_get_set_methods(self, klass, out):
...
-                    self._write_trait(out, name, t_def, vtk_set_meth,
-                                      mapped=False)
+
+                    t_def = ('traits.String({0!r}, '
+                             'enter_set=True, auto_set=False)'.format(default))
+                    self._write_trait(out, name, t_def, vtk_set_meth, mapped=False)
+
+                elif PY_VER < 3 and typ is unicode:
+                    if default == u'\x00':
+                        default = ''
```
rkern: `default = u''`

Issue #219, opened 2015-07-15
(VTK 5.8 only) VTKFileReader unable to open vtk file with unstructured grid


dmsurti:
Refer to #213 which is fixed in #217. 

The test `test_vtk_file_reader.test_unstructured_grid_file` fails on Travis which uses VTK 5.8. The sample file UGridVTK.ext has been added, but it fails with:
`vtkDataSetReader (0x12aa7b20): Unable to open file: /home/travis/build/enthought/mayavi/mayavi/tests/data/UGridEx.vtk`

stefanoborini:
Are we still supporting VTK 5? If it's being phased out, maybe we should close this one.

kitchoi:
The filename is actually `/home/travis/build/enthought/mayavi/mayavi/tests/data/uGridEx.vtk`  (note the lower case `u`).  I bet the test would have passed with vtk 5.8 if the filename is corrected.  See https://github.com/enthought/mayavi/pull/353

stefanoborini:
@kitchoi ok, then it's no longer relevant. Closing.

kitchoi:
Probably good to try un-skipping the test and see if the test passes.

stefanoborini:
@kitchoi trying now.

PR #356, opened 2016-04-25
Check if test successful with the appropriate filename

stefanoborini:
Verifies, and if green, solves #219


stefanoborini:
codecov is in partial system outage. It will take a while before being green.

stefanoborini:
Considering it's one test addition and every CI is green, merging.

Issue #336, opened 2016-04-06
array_ext.so needs to be rebuilt upon upgrading to vtk7


kitchoi:
Regarding `python3 setup.py develop`, if `tvtk/array_ext.so` already exists and is built from vtk6.x, it needs to be removed and rebuilt for vtk7.

stefanoborini:
Probably a good idea just to remove it before every build?

kitchoi:
yep, probably similar to rebuilding `tvtk_classes.zip`

stefanoborini:
Working on it.

stefanoborini:
reopening. I might have accidentally closed it.

stefanoborini:
After discussion with @cournape we decided it's probably a better idea not to go down this path. You should just clean. Mayavi is doing rather non-standard things with distutils, and we are probably trying to solve the wrong problem.

Closing as wontfix. Just clean instead.

stefanoborini:
@kitchoi pr open.

PR #358, opened 2016-04-26
Removes array_ext.so/pyd on clean

stefanoborini:
Removes the extension as from issue #336.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -386,9 +386,6 @@ class MyClean(clean.clean):
         clean.clean.run(self)
 
 
-
-
-
 # Configure our extensions to Python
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -366,6 +378,13 @@ def run(self):
         install_data.install_data.run(self)
 
 
+class MyClean(clean.clean):
+    def run(self):
+        print("Cleaning extensions")
+        clear_array_ext()
```
rkern: Just implement it all in here. Your docstring and print message should note that it is deleting extensions that were built in-place in addition to the normal compile products that are cleaned.

Issue #38, opened 2012-06-23
mlab.text() draws into wrong figure when using multiple figures


mluessi:
When using multiple mlab figures, mlab.text() always draws into the last figure instead of the figure specified via the parameter or using mlab.figure(). For example, when using

```
f1 = mlab.figure()
mlab.test_contour_surf()

f2 = mlab.figure()
mlab.test_surface()

mlab.text(0.5, 0.5, 'test', figure=f1)
```

The text appears in f2 instead of f1. The same thing happens when the last line is replaced with

```
mlab.figure(f1)
mlab.text(0.5, 0.5, 'test')
```

stefanoborini:
Confirmed as still present. The provided code does not solve the issue. Investigating.

stefanoborini:
After investigation, the problem is the following. 

The PipeFactory (in its specific instance for the TextFactory), calls add_module, which in turn works on the only available engine calling engine.add_module(), which in turns calls add_filter() which in turn wants an object of the pipeline. If an object is not present, it uses the current one.
The problem occurs because the current one is the second Surface we create.

So now I am in a conundrum. Do we have this problem because there's only one engine (while there should actually be two, one per each scene) or is it because we have one, but the current object should be replaced with the first surface? If the second, the problem is that we don't pass the surface object, we pass the destination scene. How are we supposed to obtain the surface object and are we allowed to replace current_object in the engine?

I can't proceed further until the proper design assumptions are clarified, I need @prabhuramachandran to weigh in.

I will leave this issue open and move to something else.

moosekaka:
I hope this issue gets fixed, I spent way too much time trying to figure out why mlab kept adding a surface object to the wrong scene (scene 2 instead of scene 1 for a two MlabSceneModel object). 
Finally saw stefanoborini's post, gave up trying to add using mlab pipeline and directly added the actor using tvtk.

PR #363, opened 2016-04-26
[WIP] Fix for incorrect drawing text in the wrong figure

stefanoborini:
Currently work in progress. Opening because I need feedback on the internals.

When done, solves #38


stefanoborini:
@prabhuramachandran in this case, I see we only have one engine operating but two scenes. I found some inconsistency when it comes to set the figure entry, so I added code on this regard, but it keeps plotting on the wrong window. I suspect it's because we only have one engine? Are we supposed to have an engine for every scene, or there's something else going on?

stefanoborini:
Closing PR waiting for developments.

Issue #359, opened 2016-04-26
Consistent test errors on MacOSX


stefanoborini:
The following test errors occur on my mac consistently on master.

```
======================================================================
ERROR: Tests if TVTK scene can be garbage collected.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/sborini/Work/github/mayavi/tvtk/tests/test_garbage_collection.py", line 26, in test_tvtk_scene
    self.check_object_garbage_collected(create_fn, close_fn)
  File "/Users/sborini/Work/github/mayavi/tvtk/tests/common.py", line 61, in check_object_garbage_collected
    self.assertTrue(object_collected[0])
IndexError: list index out of range

======================================================================
ERROR: Test if custom modules can be imported.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/sborini/Work/github/mayavi/tvtk/tests/test_tvtk.py", line 107, in test_custom
    r = tvtk.XMLDataReader()
  File "tvtk_classes/xml_data_reader.py", line 46, in __init__
    tvtk_base.TVTKBase.__init__(self, vtk.vtkXMLDataReader, obj, update, **traits)
  File "/Users/sborini/Work/github/mayavi/tvtk/tvtk_base.py", line 321, in __init__
    self._vtk_obj = klass()
TypeError: this is an abstract class and cannot be instantiated

```

stefanoborini:
I want to understand what's going on here.

PR #365, opened 2016-04-28
Skip test for interacting test with TVTKScene

stefanoborini:
After investigation, the problem arises from an interaction between
test_mlab_scene_model_with_gui and this test in a nose run.
I decided to skip the test because it is probably intrinsic in the
way things interact during the test.

See issue #359

### DIFF
--- a/tvtk/tests/test_garbage_collection.py
+++ b/tvtk/tests/test_garbage_collection.py
@@ -12,9 +12,18 @@ from tvtk.pyface.api import DecoratedScene, Scene
 from tvtk.pyface.scene_model import SceneModel
 from tvtk.tests.common import TestGarbageCollection
 
+
 class TestTVTKGarbageCollection(TestGarbageCollection):
     """ See: tvtk.tests.common.TestGarbageCollection
     """
+
+    # This test interacts with the mayavi.tests.test_garbage_collection.
+    # test_mlab_scene_model_with_gui. The most likely cause is that
+    # the instantiation of a MayaviScene (itself derived from TVTKScene) leads
+    # to a circular reference with the renderer, meaning that TVTKScene cannot
+    # be collected only through refcount alone.
+    @unittest.skip("Circular reference failure due to interaction with another "
+                   "test. (issue #359)")
     def test_tvtk_scene(self):
         """ Tests if TVTK scene can be garbage collected."""
         def create_fn():


itziakos:
I would prefer to fix the issue instead of  skipping the test

stefanoborini:
@itziakos problem is that I have no idea what kind of state is leaking, and I am not sure I want to dig into it as I might be redirected to another project any minute now. So I think I'll just leave it out.

itziakos:
> @itziakos problem is that I have no idea what kind of state is leaking, and I am not sure I want to dig into it as I might be redirected to another project any minute now. So I think I'll just leave it out.

The tests are working for windows and unix so skipping them is bad. The only fail on mac-osx bacause of a state leak in the test-suite. So if we do not have time at the moment to fix then the right thing to do is to skip them or probably declare them as know failures on mac-osx only.

stefanoborini:
@kitchoi do they fail on your OSX machine?

But I agree with you @itziakos . It however leads to the general problem that we don't really have a good, visual tool to understand what's going on.

kitchoi:
@stefanoborini  with VTK 6.3 and Python 2.7, I have different errors on OSX:
```
python -m nose.core tvtk/tests
.....................................................S...............E.E.......................
ERROR: Test if all the TVTK classes can be instantiated
  File "/Users/kit/ETS/mayavi/tvtk/tests/test_tvtk.py", line 655, in test_all_instantiable
    tvtk_klass = getattr(tvtk, tvtk_name, None)
  File "/Users/kit/ETS/mayavi/tvtk/tvtk_classes/tvtk_helper.py", line 1372, in <lambda>
    StructuredGridLIC2D = property(lambda self: get_class('StructuredGridLIC2D'))
  File "/Users/kit/ETS/mayavi/tvtk/tvtk_classes/tvtk_helper.py", line 36, in get_class
    mod = get_module(fname)
  File "/Users/kit/ETS/mayavi/tvtk/tvtk_classes/tvtk_helper.py", line 28, in get_module
    mod = __import__('tvtk.tvtk_classes.%s'%fname, globals(), locals(), [fname])
  File "/Users/kit/ETS/mayavi/tvtk/tvtk_classes/structured_grid_lic2d.py", line 45
SyntaxError: Non-ASCII character '\xe2' in file /Users/kit/ETS/mayavi/tvtk/tvtk_classes/structured_grid_lic2d.py on line 46, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details

ERROR: test_unicode_traits (tvtk.tests.test_tvtk.TestTVTKModule)
  File "/Users/kit/ETS/mayavi/tvtk/tests/test_tvtk.py", line 686, in test_unicode_traits
    self.assertIsInstance(reader.unicode_record_delimiters, unicode)
AttributeError: 'DelimitedTextReader' object has no attribute 'unicode_record_delimiters'

```
Doesn't seem relevant and the original errors you have did not occur for me.  I assume you are using VTK 6.3 too?

kitchoi:
Update: my tvtk_classes wasn't cleaned up.  Rebuilt and no error for me on Mac OSX.

kitchoi:
So no error with `enthought/mayavi$ python -m nose.core tvtk/tests`  (This is how .travis.yml is at the moment)
but same errors occur with `enthought/mayavi$ python -m nose`

stefanoborini:
closing this PR. Decided to investigate further.

PR #369, opened 2016-04-29
Fix/368 integrationtests fail with exit 1

### LINKED ISSUE #368: Several integrationtests fail without exit 1
Examples: https://travis-ci.org/enthought/mayavi/jobs/126611791#L1401

One needs to mimick the behaviour of integrationtests.common https://github.com/enthought/mayavi/blob/master/integrationtests/mayavi/common.py#L566

My bad, will fix this soon.

kitchoi:
Fix #368 so that failed tests are properly reported.

Fortunately, only one test from test_texture whose failure was not captured and I believe it was because I set the check criteria (a standard deviation of pixels) just a bit too high (200 versus ~180 for 28000 sample points).  However I could not reproduce the small difference in the standard deviation on my local machine (OSX or Ubuntu 12) and was not able to understand it.

### DIFF
--- a/integrationtests/mayavi/test_mlab_savefig.py
+++ b/integrationtests/mayavi/test_mlab_savefig.py
@@ -1,5 +1,6 @@
 import os
 import shutil
+import sys
 import unittest
 import tempfile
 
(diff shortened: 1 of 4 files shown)


kitchoi:
Branch is flagged red because changes are in the integrationtests and codecov found nothing to diff... :(

Issue #357, opened 2016-04-26
tvtk Window Position is of wrong type on initialisation (VTK 6.3)


kitchoi:
Tested VTK 5.8, VTK 6.3 and VTK 7.0 on Linux with Python 2.7
The following occurs only for VTK 6.3
```
>>> from tvtk.api import tvtk
>>> window = tvtk.GenericOpenGLRenderWindow()
---------------------------------------------------------------------------
TraitError                                Traceback (most recent call last)

...[skip some lines, see full traceback in the following link]...

<ipython-input-7-405e09e991ce> in <module>()
----> 1 window = tvtk.GenericOpenGLRenderWindow()

TraitError: The 'position' trait of a GenericOpenGLRenderWindow instance must be a string, but a value of (0, 0) <type 'tuple'> was specified.
```
See complete traceback here, as is captured by tvtk nose test:
https://travis-ci.org/enthought/mayavi/jobs/124533392#L863

After some investigation, this is what caused it:
```
>>> import vtk  # VTK 6.3 from vtk.org
>>> import tvtk.vtk_parser
>>> parser = tvtk.vtk_parser.VTKMethodParser()
>>> instance = parser._get_instance(vtk.vtkWindow)
>>> instance = GetPosition()
'_0000000001f02788_p_void'
```
Whilst with VTK 5.8 and VTK 7.0 (Python2.7)
```
>>> import vtk  # VTK 7.0 from vtk.org or VTK 5.8 from ubuntu
>>> import tvtk.vtk_parser
>>> parser = tvtk.vtk_parser.VTKMethodParser()
>>> instance = parser._get_instance(vtk.vtkWindow)
>>> instance = GetPosition()
(0, 0)
```

kitchoi:
Update: problem with VTK 6.2 as well
This is, to me, a vtk bug that got fixed in VTK 7.0

Issue #360, opened 2016-04-26
Occasional prop_color_value set failure on MacOS


stefanoborini:
The following tests occasionally fail on my mac. 

```
======================================================================
FAIL: Test if all the TVTK classes can be instantiated
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/sborini/Work/github/mayavi/tvtk/tests/test_tvtk.py", line 669, in test_all_instantiable
    raise AssertionError(message.format(''.join(errors)))
AssertionError: Not all classes could be instantiated:
Traceback (most recent call last):
  File "/Users/sborini/Work/github/mayavi/tvtk/tests/test_tvtk.py", line 664, in test_all_instantiable
    tvtk_klass()
  File "tvtk_classes/hardware_selector.py", line 81, in __init__
    tvtk_base.TVTKBase.__init__(self, vtk.vtkHardwareSelector, obj, update, **traits)
  File "/Users/sborini/Work/github/mayavi/tvtk/tvtk_base.py", line 334, in __init__
    self.update_traits()
  File "/Users/sborini/Work/github/mayavi/tvtk/tvtk_base.py", line 482, in update_traits
    setattr(self, name, val)
  File "/Users/sborini/.jaguar/envs/jaguar/lib/python2.7/site-packages/traits/trait_handlers.py", line 172, in error
    value )
TraitError: The 'prop_color_value' trait of a HardwareSelector instance must be a tuple of the form: (0.0 <= a floating point number <= 1.0, 0.0 <= a floating point number <= 1.0, 0.0 <= a floating point number <= 1.0), but a value of (1.0720514219217772e-35, 1.5052411792145747e-39, 9218721025359872.0) <type 'tuple'> was specified.
Traceback (most recent call last):
  File "/Users/sborini/Work/github/mayavi/tvtk/tests/test_tvtk.py", line 664, in test_all_instantiable
    tvtk_klass()
  File "tvtk_classes/open_gl_hardware_selector.py", line 44, in __init__
    tvtk_base.TVTKBase.__init__(self, vtk.vtkOpenGLHardwareSelector, obj, update, **traits)
  File "/Users/sborini/Work/github/mayavi/tvtk/tvtk_base.py", line 334, in __init__
    self.update_traits()
  File "/Users/sborini/Work/github/mayavi/tvtk/tvtk_base.py", line 482, in update_traits
    setattr(self, name, val)
[...]

PR #370, opened 2016-04-29
Fix "_p_void" values in TVTK with VTK 6.x

kitchoi:
I have tested this with VTK 6.2 and 6.3 on Ubuntu 12, and VTK 6.3 on OSX 10.11

Things seem to work but since this is a fundamental change, I would prefer sitting on it for a bit longer.  Review and comments are very much appreciated.

Most of these changes are tested by the existing TVTK test for initialising all classes, but I think I could also add more tests to make it more robust.

To compare the bahaviour, try the following with the master branch and with this one:
```
from tvtk.api import tvtk
obj = tvtk.ViewTheme()
obj.configure_traits()
```


kitchoi:
Should fix #357 and #360

tvtk/tvtk_base.py
```diff
@@ -272,6 +274,11 @@ class TVTKBase(traits.HasStrictTraits):
     # The wrapped VTK object.
     _vtk_obj = traits.Trait(None, None, vtk.vtkObjectBase())
 
+    # Stores the names of the traits whose VTK Get methods may return
+    # invalid values (e.g. reference to a point) or uninitialised values
+    # We would try to update but allow it to failure
```
prabhuramachandran: "allow it to fail" :)

tvtk/wrapper_gen.py
```diff
@@ -14,18 +14,149 @@
...
+
+    elif type_ is bool:
+        return 'traits.Bool'
+
+    else:
+        raise TypeError("Could not understand type: {}".format(type_))
+
+
+def get_array_meta(value):
+    if not isinstance(value, Sequence):
```
kitchoi: self-review: need docstring here

stefanoborini:
This is too much code to review without being told more.

kitchoi:
If you try to run the new tvtk test case with the master branch + VTK 6.3, you should see a bunch of failures: https://gist.github.com/kitchoi/8e92ad08e44334133360536ce3d5bdff
These are what this PR is fixing, plus a bunch of refactoring.

stefanoborini:
The problem now is that it's very hard to differentiate between what you changed for the fix and what you refactored.

stefanoborini:
I think it's time we ditch codecov until they fix this unexpected changes problem.

kitchoi:
If you want less of the refactoring changes, try git diff the master and this commit: https://github.com/enthought/mayavi/pull/370/commits/7a30bdd89993c38bebdfdb75212f8c758a57a300

stefanoborini:
Try pushing a dummy change. I disabled codecov

tvtk/wrapper_gen.py
```diff
@@ -511,223 +685,178 @@ def _gen_get_set_methods(self, klass, out):
...
+                        self._write_tvtk_method(klass, out, vtk_set_meth, set_sig)
+
+                    # We cannot update the trait
                     del updateable_traits[name]
 
-            else: # Has a specified range of valid values.
-                if klass.__name__ == 'vtkCubeAxesActor2D' and \
-                       name == 'inertia':
-                    # VTK bug.  Inconsistent API!
-                    rng = (float(rng[0]), float(rng[1]))
```
kitchoi: I have removed the special handling of 'vtkCubeAxesActor2D.inertia' here because from what I have researched on the VTK API, `inertia` is indeed an `int` and there is no inconsistency.

stefanoborini:
apparently the codecov yaml file did not solve for this branch

kitchoi:
I think the "codecov/project" captures the increase in coverage after we added VTK 6.3 and VTK 7 in the test matrix, it raises a red flag because it did not expect those changes (who knows how they decide what they should expect).

kitchoi:
That "unexpected changes by codecov", I guess, has to do with the fact that this PR started from a master before VTK 6.3 and VTK 7 were added to travis.  So I guess this behaviour is only temporary, i.e. future PRs should not have this problem.
Parallel to this, if we all agree that the "codecov/changes" check is not required, we can raise a new issue and fix can be done by adding `status:  changes: enabled: no` in the codcov.yml (see https://github.com/codecov/support/wiki/codecov.yml)

stefanoborini:
Exactly, I open a PR to remove the change. I'll test it on my PR for OSX and then you can merge it here if it works.

tvtk/tests/test_tvtk.py
```diff
@@ -647,34 +842,179 @@ def tearDown(self):
...
+                    trait = getattr(obj, trait_name)
+                except (TypeError, TraitError):
+                    errors_getting_trait.append(
+                        (tvtk_klass_name, trait_name))
+                else:
+                    if isinstance(trait, str) and trait.endswith('_p_void'):
+                        errors_trait_is_ptr.append(
+                            (tvtk_klass_name, trait_name, trait))
+
+        if errors_trait_is_ptr:
```
stefanoborini: if errors_getting_trait

tvtk/tests/test_tvtk.py
```diff
@@ -647,34 +842,179 @@ def tearDown(self):
...
+            message = ('These traits cannot be obtained:\n'
+                       '\n'.join(('{0}.{1}'.format(*error)
+                                  for error in errors_getting_trait)))
+        if errors_trait_is_ptr:
+            message = ('These traits are invalid:\n'
+                       '\n'.join(('{0}.{1} = {2}'.format(*error)
+                                  for error in errors_trait_is_ptr)))
+
+    def test_no_trait_has_ptr_address_as_value(self):
```
stefanoborini: Seems this method and the above have the same name.

tvtk/tests/test_tvtk.py
```diff
@@ -647,34 +842,179 @@ def tearDown(self):
...
+                        errors_trait_is_ptr.append(
+                            (tvtk_klass_name, trait_name, trait))
+
+        if errors_trait_is_ptr:
+            message = 'These traits are invalid:\n'
+            message += '\n'.join(('tvtk.{0}.{1} = {2!r}'.format(*error)
+                                  for error in errors_trait_is_ptr))
+            self.fail(message)
+
+    def test_all_traits_can_be_obtained(self):
```
stefanoborini: Is this one redundant with two methods above?

tvtk/tvtk_base.py
```diff
@@ -471,15 +478,28 @@ def update_traits(self, obj=None, event=None):
         vtk.vtkObject.GlobalWarningDisplayOff()
 
         for name, getter in self._updateable_traits_:
+            if name == 'global_warning_display':
```
stefanoborini: what about this one ? what makes it special?

tvtk/vtk_parser.py
```diff
@@ -295,7 +294,8 @@ def get_other_methods(self):
         """
         return self.other_meths
 
-    def get_method_signature(self, method):
+    @staticmethod
+    def get_method_signature(method):
```
stefanoborini: function?

tvtk/wrapper_gen.py
```diff
@@ -8,24 +8,185 @@
...
+from .common import get_tvtk_name, camel2enthought, is_version_58
+
 from . import vtk_parser
 from . import indenter
 from . import special_gen
 
 PY_VER = sys.version_info[0]
 
 
+def get_trait_def(value, kwargs="enter_set=True, auto_set=False"):
```
stefanoborini: this is a bit weird... when i see kwargs I immediately assume dict.

tvtk/wrapper_gen.py
```diff
@@ -8,24 +8,185 @@
...
+    """ Return the appropriate trait type, reformatted string and
+    the associated traits meta data for a given `value`
+
+    If a sequence is given, traits.Array is returned instead of
+    traits.Tuple or traits.List
+
+    Parameters
+    ----------
+    value
```
stefanoborini: ?

tvtk/wrapper_gen.py
```diff
@@ -8,24 +8,185 @@
...
+        except TypeError:  # Unhashable
+            continue
+
+        if all_same_type and arg_format[0] in default_mappings:
+            # All types in `arg_format` are the same and they are
+            # in the mapping (e.g. arg_format = ('int', 'int')
+            default = default_mappings[arg_format[0]]
+
+            if len(arg_format) > 1:
+                return (default,)*len(arg_format)
```
stefanoborini: How does it react to something having (int, float)?

tvtk/tests/test_tvtk.py
kitchoi: Oops, my copypasta, well spotted.

kitchoi: The first `test_no_trait_has_ptr_address_as_value` is a ghost copypasta that was not deleted, deleting it now.

tvtk/tvtk_base.py
kitchoi: This one is not originally my code, I just moved it out of the try block.  So I don't know.

tvtk/wrapper_gen.py
```diff
@@ -1336,3 +1470,251 @@ def %(setter)s(self, arg):
...
+            True, True, '_write_hardware_selector_prop_color_value'),
+
+        # In VTK 5.8, tolerance is initialised as 0 while the range
+        # is 1-100
+        'vtkAxesTransformRepresentation.Tolerance$': (
+            True, True, '_write_axes_transform_representation_tolerance')
+    }
+
+    @classmethod
+    def _get_special_matched_key(cls, klass, vtk_attr_name):
```
stefanoborini: I generally don't mind docstrings for methods that are self-descriptive, such as the one above, but this one probably needs explanation.

tvtk/wrapper_gen.py
```diff
@@ -1336,3 +1470,251 @@ def %(setter)s(self, arg):
...
+        else:
+            return None
+
+    @classmethod
+    def _get_special_updateable_failable(cls, klass, vtk_attr_name):
+        key = cls._get_special_matched_key(klass, vtk_attr_name)
+        return cls.special_traits.get(key)[:2]
+
+    @classmethod
+    def _is_special(cls, klass, vtk_attr_name):
```
stefanoborini: same

tvtk/vtk_parser.py
kitchoi: There are many occurrence of `parser = VTKMethodParser(); parser.get_method_signature(...)` in wrapper_gen and in tests (and possibly some other places I have not touched), could do with a separately cleanup PR in order to move this function out of the class?

tvtk/wrapper_gen.py
kitchoi: how about
```
value
    Can be anything
```

tvtk/wrapper_gen.py
kitchoi: `all_same_type` is False

tvtk/wrapper_gen.py
```diff
@@ -1336,3 +1470,251 @@ def %(setter)s(self, arg):
...
+        # trait name
+        name = self._reform_name(vtk_attr_name)
+
+        # vtk set method
+        vtk_set_meth = getattr(klass, 'Set'+vtk_attr_name)
+
+        self._write_trait(out, name, t_def, vtk_set_meth, mapped=False)
+
+    def _write_hardware_selector_prop_color_value(self, klass, out,
+                                                  vtk_attr_name='PropColorValue'):
```
stefanoborini: Weird that you have to pass a default value.

tvtk/wrapper_gen.py
```diff
@@ -1336,3 +1470,251 @@ def %(setter)s(self, arg):
...
+        name = self._reform_name(vtk_attr_name)
+
+        # VTK set method
+        vtk_set_meth = getattr(klass, 'Set'+vtk_attr_name)
+
+        self._write_trait(out, name, t_def, vtk_set_meth, mapped=False,
+                               force_update='False')
+
+    def _write_axes_transform_representation_tolerance(self, klass, out,
+                                                       vtk_attr_name='Tolerance'):
```
stefanoborini: Same

stefanoborini:
Review ready. Merge when done (they are mostly minor).

stefanoborini:
How is this related to #907? did you skip the test?

kitchoi:
What is issue #907?

stefanoborini:
Fixed. I pasted the wrong content.

tvtk/wrapper_gen.py
kitchoi: I don't have to. But I think it acts as a form of documentation.

stefanoborini: I understand the intent, but it's deceiving. When I see that, I don't expect it to be documentation.
You already enforce the value in the content. When I see that I assume that the method can be called without any passed vtk_attr_name for some reason, and you are "fixing" that behavior with the only acceptable value, but the vtk_attr_name is always passed, only in this routine must be one and only one value, otherwise there's a programming error.

tvtk/tests/test_tvtk.py
kitchoi: removed the first one which is the actual duplicate.

tvtk/tests/test_tvtk.py
kitchoi: errors_trait_is_ptr should be correct now.

tvtk/tvtk_base.py
stefanoborini: Ok. Leave it as is. We can't know everything.

tvtk/vtk_parser.py
stefanoborini: Leave it as is. not top priority. If you want, open a issue tagged refactoring.

tvtk/wrapper_gen.py
stefanoborini: You pretty much have to put something, even trivial, I guess? if we run the docgen on it I think it will complain (do we even run the docgen on it?). Good for me.

tvtk/wrapper_gen.py
kitchoi: I have removed the default value now.  Unfortunately I have to keep the signature consistent, so `vtk_attr_name` is kept there.

tvtk/wrapper_gen.py
stefanoborini: I meant in general.... are those cases handled at all (doesn't seem so)?

kitchoi: For those cases, the vtk set method and get method are written as is.  This is exactly what `test_spider_plot_actor` in `tvtk/test_tvtk.py` is testing.

tvtk/wrapper_gen.py
```diff
@@ -511,223 +691,178 @@ def _gen_get_set_methods(self, klass, out):
...
+                    # without argument will continue to return an address to
+                    # a pointer, so there is no point in updating.
+                    # However, some VTK Get methods have multiple signatures
+                    # among which you could pass a numpy array of the right
+                    # size and the Get method would populate the array with
+                    # the value of the attribute
+                    # e.g. x = numpy.empty(9); ImageConvolve.GetKernel3x3(x)
+                    # FIXME: we could try harder in retrieving these values
+                    # in TVTKBase.update_traits
+                    del updateable_traits[name]
```
kitchoi: If I change `del updateable_traits[name]` to `allow_update_failure.add(name)`, the difference in behaviour is not only that TVTKBase.update_traits will have more updating to do while those updating will almost certainly fail with TraitError, but I will also get these traits visual on the UI.
E.g. with `del updateable_traits[name]`
del_updateable

E.g. replace with `allow_update_failure.add(name)`
allow

Question is, which one do we want.  The bahaviour on the interactive interpreter is the same.

kitchoi:
> How is this related to issue #385 ? did you skip the test?

I did not do anything about #385, the error is still there (look at VTK 6.3 and VTK 7)
https://travis-ci.org/enthought/mayavi/jobs/128037832#L1208

tvtk/wrapper_gen.py
kitchoi: Corran suggested using TextEditor for these arrays.  Unfortunately I am preoccupied at the moment.  I think I would attack this problem in a separate PR if and after this one is merged.

tvtk/tests/test_tvtk.py
```diff
@@ -636,6 +657,180 @@ def test_null_string_wrapper(self):
             ('caption', 'GetCaption') in cap._updateable_traits_, True)
         self.assertEqual('caption' in cap._full_traitnames_list_, True)
 
+    def test_spider_plot_actor(self):
+        """ Test SpiderPlotActor Get/SetAxisLabel works
+        """
+        actor = tvtk.SpiderPlotActor()
```
prabhuramachandran: It would be nice if there were a comment as to why this object is tested that would be helpful.

kitchoi:
Note to self
TODO:
- [ ] Test with Python 3

tvtk/wrapper_gen.py
prabhuramachandran: Yes, I agree this looks a bit unusual.

kitchoi: I overlooked this comment.  Will make kwargs default to a dict, and compose the string within the function.

tvtk/wrapper_gen.py
```diff
@@ -8,24 +8,191 @@
...
+
+        kwargs += ('shape={shape}, dtype={dtype}, '
+                   'value={value!r}, cols={cols}').format(
+                       shape=shape, dtype=dtype,
+                       value=value, cols=min(3, len(value)))
+
+        return 'traits.Array', '', kwargs
+
+    elif type_ is bool:
+        return 'traits.Bool', '', kwargs
```
kitchoi: note to self, return `'traits.Bool', str(value), kwargs`

tvtk/wrapper_gen.py
prabhuramachandran: Fair enough.

tvtk/wrapper_gen.py
prabhuramachandran: It is quite possible this was a bug with VTK 4.x or so which is not relevant anymore so this is fine.

tvtk/wrapper_gen.py
```diff
@@ -1336,3 +1482,291 @@ def %(setter)s(self, arg):
...
+        vtk_set_meth = getattr(klass, 'Set'+vtk_attr_name)
+
+        self._write_trait(out, name, t_def, vtk_set_meth, mapped=False)
+
+    def _write_hardware_selector_prop_color_value(self, klass, out,
+                                                  vtk_attr_name):
+
+        if vtk_attr_name != 'PropColorValue':
+            raise RuntimeError("Not sure why you ask for me! "
+                               "I only deal with PropColorValue. Panicking.")
```
prabhuramachandran: Whoa, why is this needed?

tvtk/wrapper_gen.py
```diff
@@ -1336,3 +1482,291 @@ def %(setter)s(self, arg):
...
+    def _write_any_something_file_prefix(self, klass, out, vtk_attr_name):
+        name = self._reform_name(vtk_attr_name)
+        vtk_set_meth = getattr(klass, 'Set' + vtk_attr_name)
+        t_def = 'tvtk_base.vtk_file_prefix("")'
+        self._write_trait(out, name, t_def, vtk_set_meth, mapped=False)
+
+    def _write_image_reader2_header_size(self, klass, out, vtk_attr_name):
+
+        if vtk_attr_name != 'HeaderSize':
+            raise RuntimeError("Not sure why you ask for me! "
```
prabhuramachandran: Why is this needed?

tvtk/wrapper_gen.py
kitchoi: This is to prevent this method being used with `vtk_attr_name` other than `PropColorValue`.  This may be unnecessary given the method is private and is used internally.

prabhuramachandran:
Thanks for this, it looks nice and the code looks a lot cleaner than before!  I will merge this for now.

Issue #366, opened 2016-04-29
Add OSX to travis CI run


stefanoborini:
At the moment we are not running OSX builds on travis.
This needs to be added.

stefanoborini:
It is quite a fight because of this https://github.com/Homebrew/homebrew-science/issues/3401

PR #372, opened 2016-05-03
Add OSX to travis CI

stefanoborini:
Partially solves #366. Deploys the general infrastructure.


stefanoborini:
I will move on to make the remaining builds green, but for now, this is the basic infrastructure. Will attempt to make it green in separate PR.

travis/setup_vtk_env.sh
```diff
@@ -0,0 +1,61 @@
...
+        export DYLD_LIBRARY_PATH=/usr/local/opt/vtk/lib
+    fi
+else
+    # Linux
+    if [[ "${VTK_VERSION}" == "5.8" ]]; then
+        # Do nothing. 5.8 is already on the system
+        :
+    else
+        # This is the name of the directory after unpacking
+        VTK_PYTHON=VTK-${VTK_VERSION}.0-Linux-64bit
```
kitchoi: Thanks for the refactoring!
The VTK7 tarball downloaded from vtk.org unpacks to a different folder name, and so Linux test on VTK 7 is actually using VTK 5.8 (and it tries to download the tarball since it couldn't find the unpack directory.   Line 358 of the log here https://travis-ci.org/enthought/mayavi/jobs/126768436 shows that it is using VTK 5.8)

kitchoi: Perhaps we could modify .travis.yml to pass VTK_VERSION = 7.0.0 or 6.3.0 so that we can skip the `.0` after `${VTK_VERSION}`
And perhaps the line in `.travis.yml` that print the VTK version should really exit with 1, prevent us from thinking the tests were run with the right version when it is not.
i.e.  Instead of
```
# .travis.yml
python -c "from __future__ import print_function; import vtk; print('VTK Version:', vtk.vtkVersion.GetVTKVersion())"
```
Do
```
# python script
from __future__ import print_function
import vtk
import sys
vtk_version = vtk.vtkVersion.GetVTKVersion()
desired_version = os.environ.get('VTK_VERSION')
if vtk_version != desired_version:
    print('Loaded VTK version is', vtk_version, 'but you wanted', desired_version)
    sys.exit(1)
```
my bad, i wrote that thing.

stefanoborini: > VTK 7 is actually using VTK 5.8

Was it always like this, or is it something new with this PR?

kitchoi: it is new with this PR after the refactoring.  Originally there were two separate, but very similar scriptd for loading VTK 6.3, and VTK 7.0, and the only difference between them, really, was the name of the unpacked directory.

stefanoborini: I account for that difference with the variable. Where's the problem?

stefanoborini: Ah ok, saw it... VTK vs vtkpython

stefanoborini:
Failure due to codecov hiccup. Restarting.

stefanoborini:
Closing PR due to codecov unexpected behavior. Reopened pre-merge against master in pull request #377

PR #374, opened 2016-05-03
Fix for unexpected test code

stefanoborini:
Fix for a few strange things. The dictionary is changed in place, so it must be copied.
Additionally, it's probably supposed to call tearDown, not setUp()

### DIFF
--- a/mayavi/tests/test_mlab_integration.py
+++ b/mayavi/tests/test_mlab_integration.py
@@ -233,7 +233,7 @@ class TestMlabPipeline(TestMlabNullEngine):
         if self.less_than_or_equal_to_vtk_5_10:
             super(TestMlabPipeline, self).tearDown()
         else:
-            for engine in registry.engines.keys():
+            for engine in list(registry.engines.keys()):
                 registry.unregister_engine(engine)
 
     def test_probe_data(self):
--- a/mayavi/tests/test_mlab_integration.py
+++ b/mayavi/tests/test_mlab_integration.py
@@ -211,18 +211,18 @@ class TestMlabNullEngineMisc(TestMlabNullEngine):
 ################################################################################
 class TestMlabPipeline(TestMlabNullEngine):
     """ Test the pipeline functions.
-        For vtk versions greater than 5.6 (5.10.1 onwards), widgets need
+        For vtk versions greater than 5.10.1, widgets need
         a render window interactor to be set, otherwise an error is raised.
         As such this test checks for the current VTK version and setups a real
-        engine for vtk > 5.6 and null engine otherwise.
+        engine for vtk > 5.10 and null engine otherwise.
     """
 
     def setUp(self):
         ver = tvtk.Version()
-        self.less_than_vtk_5_6 = True
+        self.less_than_or_equal_to_vtk_5_10 = True
         if ver.vtk_major_version >= 5 and ver.vtk_minor_version >= 10:
-            self.less_than_vtk_5_6 = False
-        if self.less_than_vtk_5_6:
+            self.less_than_or_equal_to_vtk_5_10 = False
+        if self.less_than_or_equal_to_vtk_5_10:
             super(TestMlabPipeline, self).setUp()
         else:
             e = Engine()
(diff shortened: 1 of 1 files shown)


kitchoi:
Can you specify which tests fail, if possible?

stefanoborini:
This one, but it only fails on OSX for some reason.

```
ERROR: Test probe_data
  File "/Users/travis/build/enthought/mayavi/mayavi/tests/test_mlab_integration.py", line 236, in tearDown
    for engine in registry.engines:
RuntimeError: dictionary changed size during iteration
```

mayavi/tests/test_mlab_integration.py
```diff
@@ -231,9 +231,9 @@ def setUp(self):
 
     def tearDown(self):
         if self.less_than_vtk_5_6:
-            super(TestMlabPipeline, self).setUp()
+            super(TestMlabPipeline, self).tearDown()
         else:
-            for engine in registry.engines:
+            for engine in registry.engines[:]:
```
kitchoi: meant `list(registry.engines)`?

stefanoborini: @kitchoi it's the same, but I just realized it's a dictionary, not a list.

kitchoi:
`less_than_vtk_5_6` is not what it looks!  it is in fact `less_than_vtk_5_10`!

mayavi/tests/test_mlab_integration.py
```diff
@@ -211,29 +211,29 @@ def test_colorbar(self):
 ################################################################################
 class TestMlabPipeline(TestMlabNullEngine):
     """ Test the pipeline functions.
-        For vtk versions greater than 5.6 (5.10.1 onwards), widgets need
+        For vtk versions greater than 5.10.1, widgets need
```
kitchoi: Is `5.10.0` a thing?

mayavi/tests/test_mlab_integration.py
```diff
@@ -211,29 +211,29 @@ def test_colorbar(self):
...
     """
 
     def setUp(self):
         ver = tvtk.Version()
-        self.less_than_vtk_5_6 = True
+        self.less_than_or_equal_to_vtk_5_10 = True
         if ver.vtk_major_version >= 5 and ver.vtk_minor_version >= 10:
-            self.less_than_vtk_5_6 = False
-        if self.less_than_vtk_5_6:
+            self.less_than_or_equal_to_vtk_5_10 = False
```
kitchoi: `less_than_vtk_5_10` since `ver.vtk_minor_version >= 10`

mayavi/tests/test_mlab_integration.py
```diff
@@ -211,29 +211,29 @@ def test_colorbar(self):
...
             mlab.set_engine(e)
 
     def tearDown(self):
-        if self.less_than_vtk_5_6:
-            super(TestMlabPipeline, self).setUp()
+        if self.less_than_or_equal_to_vtk_5_10:
+            super(TestMlabPipeline, self).tearDown()
         else:
-            for engine in registry.engines:
+            for engine in registry.engines.keys():
```
kitchoi: In Python3, keys would be a view, so I would go for `list(registry.engines)` or `list(registry.engines.keys()`.

stefanoborini: Good point.

mayavi/tests/test_mlab_integration.py
stefanoborini: I think we can safely assume that patchlevel is irrelevant (hopefully)

kitchoi: I agree.  So perhaps `5.10 onwards` is more true to the implementation in the code.

stefanoborini:
Failure seems unrelated, restarting.

kitchoi:
Assuming tests would pass with Linux + 5.10 (as it does with Mac), it looks good to me

stefanoborini:
Tested on 5.10 ubuntu, seems to work. Merging.

PR #383, opened 2016-05-05
Disable codecov changes.

stefanoborini:
Problems still present in the codecov changes, probably temporary. We disable them.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -11,6 +11,10 @@ codecov:
       # Learn more at https://codecov.io/docs#yaml_default_commit_status
       project: true
 
+  status: 
+    changes: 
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -11,6 +11,10 @@ codecov:
       # Learn more at https://codecov.io/docs#yaml_default_commit_status
       project: true
 
+  status: 
+    changes: 
+      enabled: no
+
```
kitchoi: There is already a `status` tag under `coverage` (Line 10), and `changes` should be at the same level as `coverage:status:project`, i.e. `coverage:status:changes`.

PR #390, opened 2016-06-02
[WIP] Run integration tests on qt

itziakos:
This PR updates the testing environment so that we can run the integration tests on qt4.


fbender:
Any reason why this has been closed without mergin?

itziakos:
This was a working attempt to see look into why the integration qt tests where failing on travis-ci and fix them. However it looks that there is a strange segfault when using wx and the off-screen rendering. To solve the later problem I should probably look into integrating the changes from  #346, #347 and #350. So I have been working on a branch that consolidates the various fixes and closed this PR.

PR #405, opened 2016-06-23
Debug info on code gen

itziakos:
Add basic debug support for the code_gen script

Details:
- Add a logger and log debug info while wrapping vtk classes
- Add `-v, --verbose` option on the code_gen script
- Minor cleanup

### DIFF
--- a/tvtk/__init__.py
+++ b/tvtk/__init__.py
@@ -6,6 +6,10 @@
 """
 
 from os.path import exists, join, dirname, isdir
+import logging
+
+# use a null handler when used as a library
+logging.getLogger('tvtk').addHandler(logging.NullHandler())
 
 # The tvtk wrapper code is all typically inside one zip file.  We try to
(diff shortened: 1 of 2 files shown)


tvtk/__init__.py
```diff
@@ -6,6 +6,10 @@
 """
 
 from os.path import exists, join, dirname, isdir
+import logging
+
+# use a null handler when used as a library
+logging.getLogger('tvtk').addHandler(logging.NullHandler())
```
prabhuramachandran: Could you please explain why this is needed?

itziakos: Adding a null handler is suggested for libraries that use the python logging framework so that there is always a Handler  (see https://docs.python.org/2/howto/logging.html#configuring-logging-for-a-library).

prabhuramachandran: Thanks, do you think this link could be put in a comment above this?  Or is this obvious?

PR #410, opened 2016-07-05
Updated LUTs to include new matplotlib colorschemes

daytonb:
I had to modify the cm2lut.py script to get the cmaps which are not in
matplotlib.cm.datad. I don't understand why some of the cmaps are there
and others are in matplotlib._cm_listed.cmaps, but I figured out from
matplotlib.cm.get_cmap that they are.

Before this can really be committed, the gifs also need uploaded, but for me mlab.savefig() is just saving black png images. So I either need some direction on fixing that, or I need someone else to take my patch and upload the gifs.

All I've been able to find on the savefig bug is that I should use savefig before mlab.show() (which gen_lut_previews.py does not use) or that it's a vtk issue.


prabhuramachandran:
LGTM, thanks!  @GaelVaroquaux -- care to comment since this is mostly your code?

prabhuramachandran:
Seems to work well for me.

prabhuramachandran:
Merging for now, @GaelVaroquaux -- if there are issues they could be addressed later.

prabhuramachandran:
Thanks @daytonb!

PR #411, opened 2016-07-06
Movie support

prabhuramachandran:
Adds support for recording movies and animating timesteps.  Also added some documentation for this feature.  For a timeseries, one can simply select the "play" checkbox and see an animation.  For the timestep and `mlab.animate` one can very easily generate animations also.

### DIFF
--- a/docs/source/mayavi/tips.rst
+++ b/docs/source/mayavi/tips.rst
@@ -17,7 +17,7 @@ an animation or so and find that each time you save an image, Mayavi
 "raises" the window to make it the active window thus disrupting your
 work.  This is needed since VTK internally grabs the window to make a
 picture.  Occluding the window will also produce either blank or
-incorrect images. 
+incorrect images.
 
 If you already have a Python script, say ``script.py`` that sets up your
 visualization that you run likes so::
@@ -106,7 +106,7 @@ so:
 
   * Now run your Mayavi script.  It should run uninterrupted on this X
     server and produce your saved images.
-  
+
 This probably will have to be fine tuned to suit your taste.
 
 Many Linux systems (including Ubuntu and Debian) ship with a helper
@@ -119,15 +119,15 @@ Beware that you shouldn't call `mlab.show` or start the mainloop in the
 script, elsewhere the script will run endlessly, waiting for interaction
 in a hidden window.
 
-.. note:: 
+.. note::
 
     If you want to use Mayavi without the envisage UI or even a traits UI
     (i.e. with a pure TVTK window) and do off screen rendering with
     Python scripts you may be interested in the
     :ref:`examples_offscreen`. This simple example shows how you can use
     Mayavi without using Envisage or the Mayavi envisage application and
-    still do off screen rendering. 
-    
+    still do off screen rendering.
+
     If you are using mlab, outside of the Mayavi2 application, simply set::
 
         mlab.options.offscreen = True
@@ -210,8 +210,8 @@ case.  Here are brief instructions on how to build VTK to do this.
 
         $ export PYTHONPATH=$VTK_BUILD/bin:$VTK_BUILD/Wrapping/Python``
         $ export LD_LIBRARY_PATH=$VTK_BUILD/bin:$MESA/lib
-        
-   Now, you should be all set.  
+
+   Now, you should be all set.
 
 Once this is done you should be able to run mlab examples offscreen.
 This will work without an X display even.
@@ -226,7 +226,7 @@ default only get offscreen contexts.  If you do want a UI you will want
 to explicitly set the render window's ``off_screen_rendering`` ivar to
 ``False`` to force a mapped window.  For this reason if you might need
 to popup a full UI, it might be better to *not set*
-``VTK_USE_OFFSCREEN=ON``.  
+``VTK_USE_OFFSCREEN=ON``.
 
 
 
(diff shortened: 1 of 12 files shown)


docs/source/mayavi/tips.rst
```diff
@@ -512,13 +543,35 @@ alongside with the `visual` handy for object-movement animation, see
...
+This will create a set of images, one for each step of the animation.
+The ``movie_maker`` instance is available on each created scene.
+
+
 Animating a series of images
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Let's say you have a stack of PNG or JPEG files that are numbered
 serially that you want to animate on a Mayavi scene.  Here is a simple
 script (called ``img_movie.py``)::
```
kitchoi: I am trying to test this branch locally.  Is this img_movie.py in the repository?

tvtk/pyface/movie_maker.py
```diff
@@ -0,0 +1,92 @@
...
+    filename = Str('anim%05d.png')
+    anti_alias = Bool(True, desc='if the saved images should be anti-aliased')
+
+    ##################
+    # Private traits
+    _subdir = Str
+    _count = Int(0)
+
+    def default_traits_view(self):
+        from traitsui.api import Item, View
```
kitchoi: `from traitsui.api import Item, View` can go to the top?

tvtk/pyface/tvtk_scene.py
```diff
@@ -917,6 +922,11 @@ def _recorder_changed(self, r):
             messenger.disconnect(i_vtk, 'EndInteractionEvent',
                                  self._record_camera_position)
 
+    def _movie_maker_default(self):
+        from tvtk.pyface.movie_maker import MovieMaker
```
kitchoi: Is this to avoid import cycle?

kitchoi:
@prabhuramachandran The feature is cool!  I have a few minor comments above and one question about the `mlab.animate(support_movie=True)`:

I tested it locally and manually with an animation function that merely modify the camera.  It does not seem to make a difference whether `support_movie` is True or not that both the movie feature and the animation feature can be supported at the same time.  Or have I missed something?  Can you explain a bit more on this please?

The rest looks good to me.

mayavi/core/file_data_source.py
```diff
@@ -98,15 +133,39 @@ class FileDataSource(Source):
...
+                                   ),
+                              ),
+                              Item(name='sync_timestep'),
+                              HGroup(
+                                  Item(name='play'),
+                                  Item(name='play_delay',
+                                       label='Delay'),
+                                  Item(name='loop'),
+                              ),
+                              enabled_when='len(object.file_list) > 1'
```
prabhuramachandran: Yes, I agree, I will change this.

tvtk/pyface/movie_maker.py
prabhuramachandran: The only reason I put it there was to eliminate any UI related imports in the case of offscreen usage or with `ETS_TOOLKIT=null`.

tvtk/pyface/tvtk_scene.py
prabhuramachandran: Yes.

prabhuramachandran:
@kitchoi -- thanks for the review!  The `mlab.animate(support_movie=False)` case will not invoke the animation start, step and stop methods. I only added this for the rare case when a user explicitly does not want to call those.  This can happen if you want to run two animations simultaneously, for example (I have no idea if anyone would do it), let us say you have an animation that keeps rotating the camera about the "z" axis.  Now let us say the user also has a timeseries and hits play.  If support_movie were True, there would be a problem.  In this case the user could use support_movie=False for the animation and the timeseries steps would be recorded as a movie.  Is that clear?

mayavi/core/file_data_source.py
prabhuramachandran: Ahh, I remember why I did this.  I thought that if you start with a single file and then add a new one, it will not show the UI elements with visible_when. It appears that I am wrong.  I just tested and this does work so I will use `visible_when`.

tvtk/pyface/movie_maker.py
kitchoi: I see.  Thanks.

prabhuramachandran: It may not make a difference but I figured it was better to be safe.

prabhuramachandran:
@kitchoi -- can this be merged?

kitchoi:
> This can happen if you want to run two animations simultaneously, for example (I have no idea if anyone would do it), let us say you have an animation that keeps rotating the camera about the "z" axis. Now let us say the user also has a timeseries and hits play. If support_movie were True, there would be a problem.

I'm sorry I should have explained what I tested more clearly.  In fact I tested exactly this scenario and I have not encountered a problem.  The experience is the same for both `support_movie=True` and `support_movie=False`.  I tried starting the animation while the movie-play is on and vice versa.
I tested using traitsui 5.1.0, pyface 5.1.0 + (pyside or wxpython)

kitchoi:
Thanks @prabhuramachandran   Now I understood what the 'support_movie' does. I guess it is because of the context of the PR that I was confused.  For others using 'animate', it should be clear what 'support_movie' does. 

Please feel free to merge. Thanks.

prabhuramachandran:
@kitchoi -- thanks!  Will merge this then.

PR #415, opened 2016-07-19
Notebook support

prabhuramachandran:
Adds support for displaying Mayavi images or x3d scenes on the notebook.  The x3d scenes allow one to fully interact with the scene in the notebook.


kitchoi:
Thanks @prabhuramachandran!  Works like a charm.

screen shot 2016-07-20 at 16 47 09

mayavi/tools/notebook.py
```diff
@@ -0,0 +1,145 @@
...
+        return scene_to_x3d(scene)
+
+
+def _counter():
+    count = 0
+    while True:
+        yield count
+        count += 1
+
+counter = _counter()
```
kitchoi:
````
from itertools import counter as _counter
counter = _counter()
```
does the same thing

mayavi/tools/notebook.py
```diff
@@ -0,0 +1,145 @@
...
+
+def _fix_x3d_header(x3d):
+    id = 'scene_%d'%counter.next()
```
kitchoi: Can `counter` be local to this function?

kitchoi:
On Mac, if I set backend to 'png' I get black images...but that's a separate issue #257 
screen shot 2016-07-20 at 16 59 01

kitchoi:
If the backend is not set correctly in the middle, the behaviour is also appropriate. :+1:
screen shot 2016-07-20 at 17 02 23

kitchoi:
One super minor comment on `counter`, which can be left as is.  This PR looks good to merge to me.  Thanks @prabhuramachandran

mayavi/tools/notebook.py
prabhuramachandran: Good point, will use this instead, thanks.

mayavi/tools/notebook.py
prabhuramachandran: Not sure it can in the way it is currently implemented.

prabhuramachandran:
@kitchoi -- thanks for the review and comments, as regards changing the backend, the recommended way to do it is to call `mlab.init_notebook('png')` and not directly set the `_backend` module global.  It is safe to call it multiple times.  I will update the docs about this as well.

kitchoi:
Thanks @prabhuramachandran!  Merging.

Issue #397, opened 2016-06-21
Saving of PDF file on OS X 10.11 results in a corrupted zip file being produced


ryanpepper:
With Mayavi installed through Conda, running the following script in a Jupyter Notebook:

>%gui qt
>from mayavi import mlab
>mlab.test_flow()

then clicking in the GUI to save the scene, if I set the file output as "snapshot.pdf", I get an error message in the terminal that opened the notebook:

> 2016-06-21 12:03:43.383 python[44385:6652218] modalSession has been exited prematurely - check for a reentrant call to endModalSession:

and no PDF file is created - a file called snapshot.pdf.gz is left.

The environment I'm working in his here:
[env.txt](https://github.com/enthought/mayavi/files/325385/env.txt)

itziakos:
There are a number of modules that are involved in this procedure and it is not clear from the error output what it the reason. 

Can you reproduce the issue without using the notebook from a normal ipython prompt?

You should also run the tests (it might help to narrow down the issue)

```
# probably need to run these from site-packages
python -um nose.core -v tvtk
python -um nose.core -v mayavi
```

ryanpepper:
Hi again,

When running the tvtk tests, I have the following error:
>======================================================================
ERROR: Test if all the attributes with MinValue/MaxValue are traits
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/ryan/anaconda3/envs/mayavi/lib/python2.7/site-packages/tvtk/tests/test_tvtk.py", line 917, in test_trait_with_range
    vtk_attr_name)
  File "/Users/ryan/anaconda3/envs/mayavi/lib/python2.7/site-packages/tvtk/tests/test_tvtk.py", line 899, in get_min_max_value
    return (getattr(vtk_klass(), get_min_method)(),
SystemError: error return without exception set

>----------------------------------------------------------------------
>Ran 106 tests in 30.976s

>FAILED (SKIP=3, errors=1)

itziakos:
Thanks for the info, 

Please not that at the moment I do not have a mac os availabe for testing (I am not a mac user), so if you can try a few more things for me will help me to debug the issue later.

> Test if all the attributes with MinValue/MaxValue are traits

Although it is not good to fail tests, I do not think that this error is related to your problem.

> Trying it from jupyter console I get a slightly different error:

The error about OpenGL is wierd and it is not clear where it is coming from (no traceback). 
[...]

Issue #387, opened 2016-05-23
Error when using tvtk.visual.ring


hamogu:
The following code fails for me:
```
Python 2.7.11 |Anaconda 2.2.0 (x86_64)| (default, Dec  6 2015, 18:57:58) 
Type "copyright", "credits" or "license" for more information.

IPython 4.0.3 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: from tvtk.tools import visual

In [2]: r = visual.ring()
ERROR: In -------src-dir--------/VTK5.10.1/Filtering/vtkDemandDrivenPipeline.cxx, line 727
vtkStreamingDemandDrivenPipeline (0x112cf3840): Input port 0 of algorithm vtkPolyDataNormals(0x117f125b0) has 0 connections but is not optional.

ERROR: In -------src-dir--------/VTK5.10.1/Filtering/vtkDemandDrivenPipeline.cxx, line 727
vtkStreamingDemandDrivenPipeline (0x11d2b93c0): Input port 0 of algorithm vtkTubeFilter(0x11d2b8210) has 0 connections but is not optional.

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-2-598a65bd376f> in <module>()
----> 1 r = visual.ring()

/Users/hamogu/anaconda/lib/python2.7/site-packages/tvtk/tools/visual.pyc in __init__(self, **traits)
    771 
    772         self._create_points()
--> 773         self._pos_changed(numpy.array([0.0, 0.0, 0.0]), self.pos)
    774         self._x_changed(self.x)
    775         self._y_changed(self.y)

/Users/hamogu/anaconda/lib/python2.7/site-packages/tvtk/tools/visual.pyc in _pos_changed(self, old, new)
    867         self.normals.update()
    868         self.tube.update()
--> 869         self.actor.mapper.update()
    870         self.render()
    871 

AttributeError: 'NoneType' object has no attribute 'update'
```

I use mayavi+tvtk 4.4.0 and vtk 5.10.1 installed through anaconda on MacOSX (see code snippet for python and Ipython versions). 
I get the same error for `visual.sphere()` and  `visual.arrow()` but ` visual.box()` works with no problems.

prabhuramachandran:
[...]

PR #416, opened 2016-07-26
Fix bugs

prabhuramachandran:
Fixes a silly error, issues #397, #387, add tests for visual, fix problem with picker ui not showing, and fix a hidden execfile.

### DIFF
--- a/tvtk/pyface/tvtk_scene.py
+++ b/tvtk/pyface/tvtk_scene.py
@@ -13,6 +13,7 @@ functionality. See the class docs for more details.
 
 from __future__ import print_function
 
+import os
 import os.path
 
 from apptools.persistence import state_pickler
@@ -683,6 +684,10 @@ class TVTKScene(HasPrivateTraits):
                 self._exporter_write(ex)
             else:
                 ex.write()
+            # Work around for a bug in VTK where it saves the file as a
+            # .pdf.gz when the file is really a PDF file.
+            if f_ext == '.pdf' and os.path.exists(f_prefix + '.pdf.gz'):
+                os.rename(f_prefix + '.pdf.gz', file_name)
 
     def save_x3d(self, file_name):
         """Save scene to an X3D file (http://www.web3d.org/x3d/).
--- a/tvtk/pyface/ui/qt4/decorated_scene.py
+++ b/tvtk/pyface/ui/qt4/decorated_scene.py
@@ -4,7 +4,7 @@ etc.
 """
 # Authors: Prabhu Ramachandran <prabhu_r@users.sf.net>,
 #          Dave Peterson <dpeterson@enthought.com>
-# Copyright (c) 2006, Enthought, Inc.
+# Copyright (c) 2006-2016, Enthought, Inc.
 # License: BSD Style.
 
 # System imports.
@@ -14,13 +14,13 @@ import os
 from pyface.qt import QtGui
 
 # Enthought library imports.
-from pyface.api import ImageResource, FileDialog, OK
+from pyface.api import ImageResource
 from pyface.action.api import ToolBarManager, Group, Action
 from tvtk.api import tvtk
 from traits.api import Instance, false, Either, List
 
 # Local imports.
-from .scene import Scene
+from .scene import Scene, popup_save
 
 
 ###########################################################################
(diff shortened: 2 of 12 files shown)


tvtk/pyface/tvtk_scene.py
```diff
@@ -683,6 +684,10 @@ def save_gl2ps(self, file_name, exp=None):
                 self._exporter_write(ex)
             else:
                 ex.write()
+            # Work around for a bug in VTK where it saves the file as a
+            # .pdf.gz when the file is really a PDF file.
```
mdickinson: Is this bug known to the VTK developers? If so, it would be useful to have a reference to a VTK issue, so that future readers of this code have a way to determine whether the upstream bug has been fixed or not.

tvtk/pyface/ui/qt4/decorated_scene.py
```diff
@@ -158,6 +158,7 @@ def _save_snapshot(self):
                             "Geomview", "VRML", "Wavefront", "Open Inventor",
                             "Povray", "X3D"]
 
+            wildcard = ""
```
mdickinson: Why didn't tests catch this bug? Is there an opportunity to add a unit test here?

mdickinson:
Any chance of unit tests or regression tests for the 'visual' issue fixes?

tvtk/pyface/tvtk_scene.py
prabhuramachandran: I don't know and its hard for me to find the time to follow up but I will try to do that.

tvtk/pyface/ui/qt4/decorated_scene.py
prabhuramachandran: Because it is triggered only when you click on "save" and I am not sure the best way to test that.  Do you have any suggestions of doing this easily without popping up a UI?

prabhuramachandran:
@mdickinson -- Thanks, that is a good idea.  visual is hardly used so has been badly neglected (as can be seen).  I should invest a bit of time writing a few proper tests for it and will do so.

tvtk/pyface/ui/qt4/decorated_scene.py
mdickinson: The easiest way would probably be to use the `mock` package to mock out the `FileDialog.open` method to always return the string `'OK'`.

prabhuramachandran: Of course, but the method is private too so it smells that I have to invoke `_save_snapshot` but anyway, better a simple test than none.  Testing UI specific stuff always leaves me with a bad taste and I am never sure if it is a good use of one's time.  I wish there were easier tooling to make this easier to do or write the code in a way that we could focus on other things.  For example, there are some scenarios that are painful to test, for example if you click on a traits UI color selector from within IPython with %gui=qt it disappears -- how is one supposed to be able to test this?  For example try this:

```
mlab.test_plot3d()
```

Now click on the last (gear) icon on the toolbar and try changing the background color, the UI appears and disappears.  How should we test this?  Its frustrating that there aren't easier ways to ensure that these do not happen due to changes in traitsui/qt/pyface/whatever.

mdickinson: I agree that accessing `_save_snapshot` directory would be a smell, but I  don't think it's necessary: you should be able to call `perform` on the related `Action` object instead. You'd have to write logic to retrieve the relevant action in the first place from `DecoratedScene.actions`, and I'd suggest adding an `id` to the individual actions to make that easier.

mdickinson: And yes, I absolutely think that testing UI interactions is a good use of time: it's an essential for a professionally developed application that involves those UI interactions. There are some tools in Traits and PyFace that make testing of UI interactions somewhat easier.

prabhuramachandran: Some pointers to documentation on these tools would be useful to have and use -- in fact it would be nice to have these documented as part of the traits manual.  It has been a decade (2005?) since I read any traits/pyface documentation, so pardon my ignorance but I think the documentation on suggested patterns for this is scarce.  Thanks!

tvtk/pyface/tvtk_scene.py
prabhuramachandran: I looked at the sources and it seems the GL2PS code has changed considerably with VTK 7.0 The GL2PS exporter has been completely changed and looks like the bug shouldn't be there -- but I did not test as I do not have a VTK 7 build with GL2PS enabled on my system.

prabhuramachandran:
@mdickinson -- Coverage increased by around 4.5% with this PR thanks to a bunch of tests!  I also fixed a bunch of bugs.  I think the PR is now safe to merge and we should cut a 4.5.0 release.  Please take a look.

tvtk/tests/test_pyface_utils.py
```diff
@@ -0,0 +1,26 @@
...
+
+class TestPopupSave(unittest.TestCase):
+
+    def _make_mock_file_dialog(self):
+        m = mock.Mock(spec=FileDialog)
+        m.open.return_value = OK
+        m.path = 'mock'
+        return m
+
+    def test_popup_save(self):
```
itziakos: you could also write the symmetric test method where 'open' does not return "OK"

tvtk/tests/test_visual.py
```diff
@@ -0,0 +1,269 @@
...
+        self.assertEqual(r.tube.radius, 0.01)
+
+        # Given
+        r = visual.ring(radius=1.0, pos=(1.0, 1.0, 1.0))
+
+        # When/Then
+        self.assertEqual(r.radius, 1.0)
+        self.assertEqual(r.thickness, 0.01)
+        self.assertEqual(r.tube.radius, 0.01)
+        self.assertTrue(np.allclose(r.pos, (1.0, 1.0, 1.0)))
```
itziakos: from numpy.testing you use 'assert_array_almost_equal'

tvtk/tools/visual.py
```diff
@@ -763,38 +760,36 @@ class Ring(HasTraits):
...
         self.property.representation = self.representation
-        show_actor(self.actor)
-        self.viewer = get_viewer()
 
+        self.viewer = get_viewer()
         self.property.on_trait_change(self.viewer.scene.render)
         self.actor.on_trait_change(self.viewer.scene.render)
         self.tube.on_trait_change(self.viewer.scene.render)
 
+        HasTraits.__init__(self, **traits)
```
itziakos: why not use "super"?

tvtk/tests/test_visual.py
prabhuramachandran: That one slipped, I have used it right below...  Is it preferred to use `assert_array_almost_equal` or `assert_allclose`??

tvtk/tools/visual.py
prabhuramachandran: No particular reason.  I wrote the tests and got it all working and fixed the issues.  This can be done later if someone cares or has the time.

prabhuramachandran:
@itziakos -- I've fixed the issues you pointed out and added another test for the visual.frame object.

itziakos:
@prabhuramachandran from a quick look it looks ok, especially when it improves coverage.

PR #417, opened 2016-07-28
Notebook Python 3 fixes

patricksnape:
- Use `next()` rather than `.next()` which doesn't exist on Python 3.
  - Use `base64` module for `png` backend.


prabhuramachandran:
Thanks for this!  Merging.

PR #421, opened 2016-08-01
Update docs for 4.5.0.

prabhuramachandran:
Also clean up some junk commits from years ago which seem to have crept
in.


prabhuramachandran:
Self-merging to get latest docs live.

PR #471, opened 2016-12-20
Adding information/tip about figure width

habi:
Because it took me way too long to figure this out (I was expecting a tuple :) )

As part of http://24pullrequests.com/


prabhuramachandran:
Thank you, merging.

PR #486, opened 2017-02-09
Remove known bug about Python 3 support

jabooth:
Python 3 is now supported in Mayavi 4.5.0.


prabhuramachandran:
Thanks for the fix.

PR #498, opened 2017-04-05
avoid loading nbextension via absolute URL

minrk:
1. load x3dom CDN over HTTPS
2. never load nbextensions with an absolute path, which only works when there is no base url (won't work in JupyterHub)

### DIFF
--- a/mayavi/tools/notebook.py
+++ b/mayavi/tools/notebook.py
@@ -94,7 +94,7 @@ def scene_to_x3d(scene):
     html = '''
     %s
     <script type="text/javascript">
-    require(["%s/x3dom.js"], function(x3dom) {
+    require(["%s/x3dom"], function(x3dom) {
         var x3dom_css = document.getElementById("x3dom-css");
         if (x3dom_css === null) {
             var l = document.createElement("link");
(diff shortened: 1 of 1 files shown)


prabhuramachandran:
@minrk -- thanks, this is a busy week for me.  I will take a look at this and other pending PRs next week or more likely week after that when my teaching duties end.

kwotuveang3k4bk:
I am sorry, but I do not see how https://www.x3dom.org/download/ should work, should not it be: https://www.x3dom.org/download/1.7.2/ ?

kwotuveang3k4bk:
But the relative path also does not work for me, I get:
`/user/username/notebooks/path/to/my/notebook/and/notebook/name/nbextensions/mayavi/x3d/x3dom.js`
but it should be: 
`/user/username/nbextensions/mayavi/x3d/x3dom.js`
and previously it was: 
`/nbextensions/mayavi/x3d/x3dom.js`

Can anyone confirm this change is working with Jupyterhub?

minrk:
I was able to test with JupyterHub just now. I had to also remove the `.js` extension from the require string (require shouldn't be passed `.js`, just like Python import shouldn't get `.py`), but now it definitely works.

prabhuramachandran:
Thanks, LGTM, merging for now.

PR #507, opened 2017-05-11
Fix bug in mouse_pick_dispatcher

FedeMiorelli:
Wrong indentation level causes only first registered picker to be checked


prabhuramachandran:
Thanks, LGTM.

PR #508, opened 2017-05-11
Fix pick_world bug with VTK >= 6

FedeMiorelli:
The pick_world function is still using the old VTK pipeline functions, added a check to that effect.

### DIFF
--- a/tvtk/pyface/picker.py
+++ b/tvtk/pyface/picker.py
@@ -29,7 +29,7 @@ from tvtk.tvtk_base import TraitRevPrefixMap, false_bool_trait
 from tvtk.common import configure_input
 from apptools.persistence import state_pickler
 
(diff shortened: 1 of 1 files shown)


tvtk/pyface/picker.py
```diff
@@ -29,6 +29,7 @@
 from tvtk.common import configure_input
 from apptools.persistence import state_pickler
 
+VTK_MAJOR_VERSION = tvtk.Version().vtk_major_version
```
prabhuramachandran: You can just use `from tvtk.common import vtk_major_version` to get this instead of doing this again.

FedeMiorelli:
@prabhuramachandran OK, modified

prabhuramachandran:
Thank you!

PR #528, opened 2017-07-16
Qt5 support

prabhuramachandran:
- Basic support for Qt5.  This makes Mayavi compatible with PyQt5.
- This requires the latest traits, traitsui, and pyface from master.  A new release may be forthcoming for these packages soon.
- This works well with VTK-8.x which is typically not available on most Python packages but is easy to do by hand.  On older versions, on MacOS, the embedded VTK window is half the size.  
- There are some warnings about "Attempt to set a screen on a child window." which seem related to the toolbar used on the decorated scene.  This seems like a pyface issue with the toolbar.
- Mlab and traits based dialogs work fine.
- Mayavi2 works but reparenting windows seems to not work very well suggesting that there are more migration issues in the other libraries.
- This PR also sneaks in a couple of Python3 migration bugs in Mayavi.

### DIFF
--- a/tvtk/code_gen.py
+++ b/tvtk/code_gen.py
@@ -132,7 +132,7 @@ class TVTKGenerator:
         specified classes.
 
         """
-        # Wrappers for the ancesors are generated in order to get the
+        # Wrappers for the accesors are generated in order to get the
         # _updateable_traits_ information correctly.
         nodes = []
         for name in names:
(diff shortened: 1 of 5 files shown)


raphmur:
bump

Great work

Can we get a release soon to try and integrate this new version in conda-forge?

prabhuramachandran:
Perhaps next week, I am waiting for a traitsui and pyface release.  Not sure about conda-forge but I will push it to pypi and then mayavi will be completely pip installable with the VTK wheels I just pushed a few weeks ago.

raphmur:
brilliant
looking forward to it

adam-grant-hendry:
Linking this to [QPainter::begin: Paint device returned engine == 0, type: 1](https://github.com/pyvista/pyvistaqt/discussions/194)

PR #530, opened 2017-07-22
Addition to the code snippet

krishnaw14:
It was kind of confusing. This code snippet used different import statements than the code snippet before it.


prabhuramachandran:
Thanks, this is a good start.

PR #531, opened 2017-07-25
Added installation using edm

krishnaw14:
I have added instructions for installation via edm. If this is fine, then I will further update the installation part.

PR #533, opened 2017-07-30
Installation using conda

krishnaw14:
I have added installation using conda.


prabhuramachandran:
I think you can put this together with the other installation related changes in the same PR, just push additional commits to your original branch and they will show up instead of sending a new PR.  The same applies to #534.

krishnaw14:
Added the two installation changes in #531

prabhuramachandran:
Closing in favor of #531.

PR #544, opened 2017-08-11
Fix numpy issues in lut_manager

FedeMiorelli:
Recent numpy versions will not accept floating values for slice indices: cast to int the result of round()


prabhuramachandran:
Thanks!  Merging.

PR #555, opened 2017-10-18
FIX: Fix deprecated .set / .get use

### LINKED ISSUE #554: BUG: Traits deprecation warnings
Running a recent version of traits/traitsui/pyface/mayavi, I get a lot of warnings like this:
```
  /home/larsoner/miniconda3/envs/mne/lib/python3.6/site-packages/mayavi/tools/helper_functions.py:107: DeprecationWarning: use "HasTraits.trait_get" instead
    traits = self.get(self.class_trait_names())
  /home/larsoner/miniconda3/envs/mne/lib/python3.6/site-packages/mayavi/tools/sources.py:101: DeprecationWarning: use "HasTraits.trait_set" instead
    super(MlabSource, self).set(trait_change_notify, **traits)
  /home/larsoner/miniconda3/envs/mne/lib/python3.6/site-packages/mayavi/tools/sources.py:101: DeprecationWarning: use "HasTraits.trait_set" instead
    super(MlabSource, self).set(trait_change_notify, **traits)
  /home/larsoner/miniconda3/envs/mne/lib/python3.6/site-packages/mayavi/tools/sources.py:216: DeprecationWarning: use "HasTraits.trait_set" instead
    pd.set(points=points)
  /home/larsoner/miniconda3/envs/mne/lib/python3.6/site-packages/mayavi/tools/sources.py:217: DeprecationWarning: use "HasTraits.trait_set" instead
    pd.set(polys=polys)
  /home/larsoner/miniconda3/envs/mne/lib/python3.6/site-packages/mayavi/core/scene.py:217: DeprecationWarning: use "HasTraits.trait_set" instead
    obj.set(scene=self.scene, parent=self)
  /home/larsoner/miniconda3/envs/mne/lib/python3.6/site-packages/mayavi/modules/glyph.py:75: DeprecationWarning: use "HasTraits.trait_set" instead
    frontface_culling=False)
  /home/larsoner/miniconda3/envs/mne/lib/python3.6/site-packages/mayavi/core/lut_manager.py:225: DeprecationWarning: use "HasTraits.trait_set" instead
    width=0.8, height=0.17)
  /home/larsoner/miniconda3/envs/mne/lib/python3.6/site-packages/mayavi/core/lut_manager.py:228: DeprecationWarning: use "HasTraits.trait_set" instead
    value=(0.1, 0.01, 0.0))
  /home/larsoner/miniconda3/envs/mne/lib/python3.6/site-packages/mayavi/core/lut_manager.py:325: DeprecationWarning: use "HasTraits.trait_set" instead
    ramp='sqrt')
```
... and so on.

Is there some preferred way to change Mayavi to handle this deprecation (try/except everywhere, or put try/except inside a single helper fuction and use the helper everywhere)? If there is I can try to make a PR.

larsoner:
Closes #554.

I started replacing all of the `.set(...)` calls with `.trait_set(...)`, until I realized that (at least) the `MlabSource` class, which is the base class for a number of things, has a special / different `.set(...)` method that does some stuff before calling `self.trait_set(...)`. I suspect some of the changes here, then, are not so good because they might lead to calling `MlabSource.trait_set(...)` directly instead of `MlabSource.set(...)`.

Perhaps the correct fix is to add a `MlabSource.trait_set(...)` that just wraps to the `MlabSource.set(...)` -- then I could change all of the `.set(...)` calls to `.trait_set(...)`, but I don't know enough about the hierarchy / ramifications to know.

In the meantime, this at least gets rid of the warnings

### DIFF
--- a/mayavi/tools/pipe_base.py
+++ b/mayavi/tools/pipe_base.py
@@ -154,7 +154,7 @@ class PipeFactory(HasPrivateTraits):
             self._target.add_trait('mlab_source', Instance(ms.__class__))
             self._target.mlab_source = ms
 
-        traits = self.get(self.class_trait_names())
+        traits = self.trait_get(self.class_trait_names())
         [traits.pop(key) for key in list(traits.keys())
                                     if key[0] == '_' or key is None]
         traits.update(kwargs)
@@ -167,8 +167,8 @@ class PipeFactory(HasPrivateTraits):
     def set(self, trait_change_notify=True, **traits):
         """ Same as HasTraits.set except that notification is forced,
         unless trait_change_notify==False"""
-        HasPrivateTraits.set(self, trait_change_notify=trait_change_notify,
-                                    **traits)
+        HasPrivateTraits.trait_set(
+            self, trait_change_notify=trait_change_notify, **traits)
         if trait_change_notify == False:
             return
         for trait in traits:
--- a/mayavi/core/lut_manager.py
+++ b/mayavi/core/lut_manager.py
@@ -218,14 +218,14 @@ class LUTManager(Base):
 
         # Initialize the scalar bar.
         sc_bar = self.scalar_bar
-        sc_bar.set(lookup_table=self.lut,
-                   title=self.data_name,
-                   number_of_labels=self.number_of_labels,
-                   orientation='horizontal',
-                   width=0.8, height=0.17)
+        sc_bar.trait_set(lookup_table=self.lut,
+                         title=self.data_name,
+                         number_of_labels=self.number_of_labels,
+                         orientation='horizontal',
+                         width=0.8, height=0.17)
         pc = sc_bar.position_coordinate
-        pc.set(coordinate_system='normalized_viewport',
-               value=(0.1, 0.01, 0.0))
+        pc.trait_set(coordinate_system='normalized_viewport',
+                     value=(0.1, 0.01, 0.0))
         self._shadow_changed(self.shadow)
 
         # Initialize the lut.
@@ -240,8 +240,8 @@ class LUTManager(Base):
         ltp.on_trait_change(self.render)
 
         # Initialize the scalar_bar_widget
-        self.scalar_bar_widget.set(scalar_bar_actor=self.scalar_bar,
-                                   key_press_activation=False)
+        self.scalar_bar_widget.trait_set(scalar_bar_actor=self.scalar_bar,
+                                         key_press_activation=False)
         self._number_of_colors_changed(self.number_of_colors)
 
 
@@ -319,10 +319,10 @@ class LUTManager(Base):
                 saturation_range = 0.0, 0.0
                 value_range = 0.0, 1.0
         lut = self.lut
-        lut.set(hue_range=hue_range, saturation_range=saturation_range,
-                value_range=value_range,
-                number_of_table_values=self.number_of_colors,
-                ramp='sqrt')
+        lut.trait_set(hue_range=hue_range, saturation_range=saturation_range,
+                      value_range=value_range,
+                      number_of_table_values=self.number_of_colors,
+                      ramp='sqrt')
         lut.modified()
         lut.force_build()
 
--- a/mayavi/core/module_manager.py
+++ b/mayavi/core/module_manager.py
@@ -257,7 +257,7 @@ class ModuleManager(Base):
             obj.stop()
         # Setup and start the new ones.
         for obj in added:
-            obj.set(module_manager=self, scene=self.scene, parent=self)
+            obj.trait_set(module_manager=self, scene=self.scene, parent=self)
             if self.running:
                 # It makes sense to start children only if we are running.
                 # If not, the children will be started when we start.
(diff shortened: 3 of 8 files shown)


larsoner:
Assuming Travis is not spuriously broken, it looks like this does break some things:

https://travis-ci.org/enthought/mayavi/jobs/289747675#L1773

So perhaps wrapping `trait_set(...)` is the way to go ... ?

mayavi/tools/pipe_base.py
```diff
@@ -167,8 +167,8 @@ def __init__(self, parent, **kwargs):
     def set(self, trait_change_notify=True, **traits):
```
rkern: This can probably be renamed to `trait_set()` now. The intention was for it to override `HasTraits.set()`, which was renamed to `trait_set()`, so this override should follow suit.

larsoner:
Okay this seems to mostly work, but has broken the text rendering on one Travis build:

https://travis-ci.org/enthought/mayavi/jobs/290002108#L1735

I can't replicate locally, has anyone seen something like this?

Failure is not present in `master` because I checked in #556 and it passes.

larsoner:
The same failure was even present on the much simpler previous commit (which did not change `set`->`trait_set`):

https://travis-ci.org/enthought/mayavi/jobs/289747675#L1773

So maybe that is some hint that the first two commits are problematic somewhere.

prabhuramachandran:
@larsoner -- sorry for the delay in getting back, I'll take a closer look this weekend.  End of semester + scipy India have kept me busy.

mayavi/tools/static/x3d/x3dom.js
```diff
@@ -756,21 +756,21 @@ that._moveable.setAttribute("matrix",that._matrixTrafo.toGL().toString());}
...
...his.f()}F.prototype.f=function(){var b=this.buffer,a,c=b.length,d=new(B?Uint8Array:Array)(c<<1);if(B)d.trait_set(b);else for(a=0;a<c;++a)d[a]=b[a];return this.buffer=d};F.prototype.d=function(b,a,c){var d=this.buffer,e=this.index,f=this.m,g=d[e],k;c&&1<a&&(b=8<a?(H[b&255]<<24|H[b>>>8&255]<<16|H[b>>>16&255]<<8|H[b>>>24&255])>>32-a:H[b]>>8-a);if(8>a+f)g=g<<a|b,f+=a;else for(k=0;k<a;++k)g=g<<1|b>>a-k-1&1,8===++f&&(f=0,d[e++]=H[g],g=0,e===d.length&&(d=this.f()));d[e]=g;this.buffer=d;this.m=f;this.index=e};F.prototype.finish=function(){var b=this.buffer,a=this.index,c;0<this.m&&(b[a]<<=8-this.m,b[a]=H[b[a]],a++);B?c=b.subarray(0,a):(b.length=a,c=b);return c};var ba=new(B?Uint8Array:Array)(256),ca;for(ca=0;256>ca;++ca){for(var K=ca,da=K,ea=7,K=K>>>1;K;K>>>=1)da<<=1,da|=K&1,--ea;ba[ca]=(da<<ea&255)>>>0}var H=ba;function ja(b,a,c){var d,e="number"===typeof a?a:a=0,f="number"===typeof c?c:b.length;d=-1;for(e=f&7;e--;++a)d=d>>>8^O[(d^b[a])&255];for(e=f>>3;e--;a+=8)d=d>>>8^O[(d^b[a])&255],d=d>>>8^O[(d^b[a+1])&255],d=d>>>8^O[(d^b[a+2])&255],d=d>>>8^O[(d^b[a+3])&255],d=d>>>8^O[(d^b[a+4])&255],d=d>>>8^O[(d^b[a+5])&255],d=d>>>8^O[(d^b[a+6])&255],d=d>>>8^O[(d^b[a+7])&255];return(d^4294967295)>>>0}
```
prabhuramachandran: This is incorrect and should not be part of this diff at all.  Can you please remove this change and push the commit or rebase it?

larsoner: No idea how that got in there, but it should be fixed now. I can squash later if need be (or you can do it during eventual merge)

mayavi/tools/pipe_base.py
prabhuramachandran: For the sake of backwards compatibility, there shouldn't be any harm in retaining "set" just for the `MlabSource`?

prabhuramachandran:
Thanks, this looks much better, I am happy to merge this for now.

larsoner:
Okay but Travis and Text rendering were broken last I checked

larsoner:
See comment ~6 above this one from [oct 19](https://github.com/enthought/mayavi/pull/555#issuecomment-337964248) and the one that follows -- it passed on `master` (I opened the spelling-fix PR nearly concurrently) but failed on the first commits of this PR

prabhuramachandran:
Yes, the spelling one also broke travis so I think it has nothing to do with this PR.  The tests need a lot of attention.  I am not sure the image based tests are very helpful here -- VTK uses this approach but it does make more sense for them than us.  The image based test infrastructure is a bit too fragile with the differing versions of Qt, Wx, VTK, etc.

larsoner:
Oh weird, I thought they were different. I guess not!

PR #559, opened 2017-11-15
Prevent get_tvtk_class_names from hijacking stdout

wmvanvliet:
The function `tvtk.tools.tvtk_doc.get_tvtk_class_names` iterates over all
VTK classes and briefly instantiates them to check whether that works.
One of those classes is `vtk.vtkPythonInteractiveInterpreter`, which upon
instantiation wil hijack `sys.stdout` to redirect all output somewhere
else. This breaks for example Jupyter notebooks, who want to hijack `sys.stdout` for themselves.

This PR restores `sys.stdout` after instantiating
`vtkPythonInteractiveInterpreter`.

fixes #558, #503

PR #564, opened 2017-11-22
Prevent get_tvtk_class_names from hijacking stderr

wmvanvliet:
This is something I missed during #559. `sys.stderr` is also hijacked.


prabhuramachandran:
Thanks!

PR #566, opened 2017-11-25
Bug fix: solve the unicode encode error in windows

Sahloul:
The error `UnicodeEncodeError: 'cp932' codec can't encode character '\xe9' in position` appears when building the documentation in windows from master source, and it can be simply solved by telling the the write wrapper to encode the text as UTF-8.

### DIFF
--- a/tvtk/code_gen.py
+++ b/tvtk/code_gen.py
@@ -7,6 +7,7 @@
 
 from __future__ import print_function
 
+import io
 import vtk
 import os
 import os.path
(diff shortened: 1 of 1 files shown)


prabhuramachandran:
@Sahloul -- thanks for the fix, this will not work on Python 2.7.  Could you please fix that if possible or let me know if you want me to take a look.

Sahloul:
@prabhuramachandran If you do not mind, please do what you think is necessary.
Thank you.

PR #578, opened 2017-12-12
FIX: Remove debug print

larsoner:
This appears to be a leftover debugging statement -- no other camera actions are `print`ed so this seems unnecessary. (I also get a flood of `setting roll` calls when automating view changes.)


prabhuramachandran:
Thanks!

Issue #568, opened 2017-11-25
`set_update_extent()` is no longer available


Sahloul:
`array_source` utilizes `vtk::ImageChangeInformation`, and calls `set_update_extent` (lines 195 and 236) that does not exists.

PR #583, opened 2017-12-28
VTK 8.x compatibility fixes

FedeMiorelli:
Removed calls to get/set_update_extent which are no longer available in VTK 8.x (reported in #568)
I am not sure this is the best way to go about it, but it works for me.


prabhuramachandran:
Thanks, I am merging this for now.

PR #588, opened 2018-01-10
MRG: Avoid NumPy deprecation

larsoner:
Gets rid of these warnings:
```
  mayavi/tvtk/array_handler.py:268: FutureWarning: Conversion of the second argument of issubdtype from `complex` to `np.complexfloating` is deprecated. In future, it will be treated as `np.complex128 == np.dtype(complex).type`.
```


prabhuramachandran:
Thanks!

PR #607, opened 2018-02-15
Fix #605: Interactors now work correctly with Qt5.

### LINKED ISSUE #605: Interactors not working with VTK-8.1.0
With latest mayavi, VTK-8.1.0, and Qt5 (Python 3.6), you cannot interact with any of the interactors at all.  For example, you can't drag the scalar bar or any of the other interactors (scalar cut plane).

If you press a key we also get this exception on some key presses, unable to reproduce exactly how, am just  reporting the traceback here so I can fix it later:

  File "git/ETS/mayavi/tvtk/pyface/ui/qt4/scene.py", line 203, in keyPressEvent
    QVTKRenderWindowInteractor.keyPressEvent(self, e)
  File "git/ETS/mayavi/tvtk/pyface/ui/qt4/QVTKRenderWindowInteractor.py", line 470, in keyPressEvent
    ctrl, shift, key, 0, key_sym)
TypeError: SetEventInformationFlipY argument 5: a string of length 1 is required
Abort trap: 6

prabhuramachandran:
This was again a problem with the pixel ratio on HDPI displays.  The
problem exists upstream in VTK's QVTKRenderWindowInteractor too and will
need to be fixed there, however, Mayavi should work fine with this fix.

### DIFF
--- a/tvtk/pyface/ui/qt4/QVTKRenderWindowInteractor.py
+++ b/tvtk/pyface/ui/qt4/QVTKRenderWindowInteractor.py
@@ -207,11 +207,20 @@ class QVTKRenderWindowInteractor(QVTKRWIBaseClass):
 
         # create qt-level widget
         if QVTKRWIBase == "QWidget":
-            QWidget.__init__(self, parent)
+            if "wflags" in kw:
+                wflags = kw['wflags']
+            else:
+                wflags = Qt.WindowFlags()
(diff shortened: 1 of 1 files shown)


prabhuramachandran:
Most of the failures are issues with the mlab.volume that I will look at next.  Merging this for now and will address the build later.

PR #641, opened 2018-05-21
Use new pyface toolkit

### LINKED ISSUE #626: Pyface and TraitsUI accept "qt" as a synonym for "qt4"
We now allow "qt" as a synonym for `ETSConfig.toolkit` being set to `qt4` in Pyface and TraitsUI.  Allowing for Mayavi would be good; may require revisiting toolkit code in Mayavi, possibly to use `pkg_resources` extension points.

prabhuramachandran:
Fixes #626.

@corranwebster -- could you please take a look at this PR, the pyface toolkit related bits especially to see if I haven't done anything wrong.  Thanks!  BTW, the changes for the pyface toolkit are all in the first commit.


corranwebster:
On a quick look-over, I think you should explicitly attempt to import `pyface.toolkit` module in the `tvtk.pyface.*.init` modules to ensure that appropriate initialization of the UI occurs (see, eg. https://github.com/enthought/traitsui/blob/master/traitsui/wx/toolkit.py#L32 and https://github.com/enthought/traitsui/blob/master/traitsui/qt4/toolkit.py#L24).  In particular, this ensures that the equivalent of the code that was deleted that initializes the `_app` instance gets run somewhere.

It is suboptimal that importing `pyface.toolkit` has important side-effects, but we are stuck with that behaviour, I think (in particular, you can't make some Qt calls until the app exists).

I'll have a closer look tomorrow morning.

corranwebster:
It's probably fine, assuming that you can't use Mayavi without bringing in TraitsUI at some point in the process.

PR #660, opened 2018-06-27
MRG: Fix window size for screenshot on HiDPI

larsoner:
On macOS with HiDPI `mlab.screenshot` is broken, only rendering 1/4 of the window (as expected by the 2x logical to physical pixel ratio for the monitor):
```
>>> from mayavi import mlab
mlab.>>> mlab.test_plot3d()
<mayavi.modules.surface.Surface object at 0x128adb728>
>>> img = mlab.screenshot()
>>> import matplotlib.pyplot as plt
>>> plt.imshow(img)
<matplotlib.image.AxesImage object at 0x133546ef0>
>>> plt.show()
```
On `master`:
figure_1

On this PR:

figure_2

Not 100% sure `figure.scene.render_window.size` is the right thing to use here, but it works.

### DIFF
--- a/mayavi/tools/figure.py
+++ b/mayavi/tools/figure.py
@@ -300,7 +300,9 @@ def screenshot(figure=None, mode='rgb', antialiased=False):
     """
     if figure is None:
         figure = gcf()
-    x, y = tuple(figure.scene.get_size())
+    # don't use figure.scene.get_size() here because this will be incorrect
+    # for HiDPI systems
+    x, y = tuple(figure.scene.render_window.size)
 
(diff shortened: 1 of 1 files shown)


prabhuramachandran:
Thanks!  LGTM!

larsoner:
FWIW I just noticed `savefig` does something different involving `magnification`:

https://github.com/enthought/mayavi/blob/master/mayavi/tools/figure.py#L234

Maybe the two approaches should be unified?

prabhuramachandran:
Yes, that would be good. :)

PR #664, opened 2018-06-29
ENH: Use views in set_id_type_array_py

larsoner:
Using this snippet (modified to properly set `empty(..., int)` so the C works properly):
```
from tvtk.array_handler import set_id_type_array_py
from tvtk.array_ext import set_id_type_array
import numpy as np

n = 100000
cs = 10
a = np.arange(cs*n)
a.shape = n, cs
b = np.empty(n*(cs+1), int)

%timeit set_id_type_array(a, b)
%timeit set_id_type_array_py(a, b)
```
on `master` I get roughly ~2x slower:
```
785 µs ± 6.39 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
1.4 ms ± 8.46 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
On this PR (which only uses views which do not copy the actual data / are for the most part free) I get only ~25% slower:
```
777 µs ± 1.34 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
961 µs ± 2.18 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
So very close in speed. This makes me think compiling might not ever be worth the hassle.

FWIW in a previous version of this (@prabhuramachandran your original snippet) there was `b = np.empty(n*(cs+1))` without specifying the type as `int`. The Python code at least gave the correct/expected output (integers stored in the `float` array `b`) whereas the C code gave garbage. But I doubt this is really done / useful in practice. There should maybe also be an `assert` statement about the dtype of `out_array` at some point.

### DIFF
--- a/tvtk/array_handler.py
+++ b/tvtk/array_handler.py
@@ -76,18 +76,18 @@ def set_id_type_array_py(id_array, out_array):
         "out_array must be contiguous."
     shp = id_array.shape
     assert len(shp) == 2, "id_array must be a two dimensional array."
-    sz = numpy.size(out_array)
+    sz = out_array.size
     e_sz = shp[0]*(shp[1]+1)
     assert sz == e_sz, \
         "out_array size is incorrect, expected: %s, given: %s" % (e_sz, sz)
 
-    # numpy.insert does the trick for us albeit slower than our Cython
-    # implementation (by as much as 4x).
-    res = numpy.insert(id_array, 0, shp[1], axis=1)
-    if len(out_array.shape) > 1:
-        out_array[:] = res
-    else:
-        out_array[:] = res.ravel()
+    # we are guaranteed contiguous, so these just change the view (no copy)
+    out_shp = out_array.shape
+    out_array.shape = (shp[0], shp[1] + 1)
+    out_array[:, 0] = shp[1]
+    out_array[:, 1:] = id_array
+    out_array.shape = out_shp
+
 
 
 if not HAS_ARRAY_EXT:


prabhuramachandran:
Ahh yes, this is very nice!  This tells me nicely that I should not benchmark stuff late in the night with little sleep.  Since the compilation is now optional anyway it should be OK to just leave it as it is.  My original benchmarks were very different and did the right thing, this was something I ran last night.  Thank you for the extra pair of eyes and a nicer implementation.

larsoner:
> On my machine I still get close to 50% slowdown and with smaller arrays even 2x.

But the call time will correspondingly shrink when the array size gets smaller. Will this actually affect real-world use cases? I wonder if at this point there is extra need-to-compile induced complexity here without any real-world benefit for users.

prabhuramachandran:
The issue is not for smaller data but for larger data where a 50% reduction means more time taken.  It does affect "real" world use cases with a large number of cells.  They are not for the majority of the casual users but for actual cases with larger data.  So given that there is a small increase in performance and that everything works, I think this is fine.  There is also the possibility of other code (at Enthought for example) where this may be used and I don't want to change that.

larsoner:
FWIW I think for larger arrays the performance difference should
(asymptotically) disappear.

>

larsoner:
Ahh, so it is. Yet another lesson for me about not making (bad) assumptions about optimization!

prabhuramachandran:
No worries, the `%timeit -o` was very handy here.

PR #677, opened 2018-07-06
MAINT: replace numpy.* by np.*

### LINKED ISSUE #673: prefer np.* to numpy.*
Hello,

In the documentation (and the code as well), there is a mix of numpy imports, either as np or as numpy.
From the user point of view, the documentation that uses numpy.* is not ideal. Indeed, using the np alias is now extremely common in the scientific community (numpy, scipy, many scikits...).

I would suggest to replace all the numpy.* by np.* It the maintainers agree, I can take care of it of course.

sciunto:
Closes #673
I made the edits in the doc and at few other places until I get bored :)

### DIFF
--- a/docs/source/tvtk/README.txt
+++ b/docs/source/tvtk/README.txt
@@ -461,10 +461,10 @@ Here is a simple example demonstrating these::
 
     >>> ########################################
     >>> from tvtk.api import tvtk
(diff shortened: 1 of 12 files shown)


prabhuramachandran:
Thanks, merging!

PR #678, opened 2018-07-09
Add a module to make streamlines from an array of seeds

Cadair:
A long time ago, I needed to make streamlines from a specific set of points. This is my first attempt to graft it into mayavi. 

It would be nice if it could be integrated into the main `Streamline` class rather than having to make a new one, but I don't know how to make the `Streamline.seed` attribute be either a `SourceWidget` or `PolyData`. I guess it would be possible to make a variant on `SourceWidget` which is just a wrapper around `PolyData` but that seems a little overkill.

### DIFF
--- a/mayavi/modules/streamline.py
+++ b/mayavi/modules/streamline.py
@@ -13,7 +13,7 @@ from math import sqrt
 
 # Enthought library imports.
 from traits.api import Instance, Bool, TraitPrefixList, Trait, \
-                             Delegate, Button
+                             Delegate, Button, Array
 from traitsui.api import View, Group, Item, InstanceEditor
 from tvtk.api import tvtk
 from tvtk.common import configure_outputs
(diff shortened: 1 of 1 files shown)


mayavi/modules/streamline.py
```diff
@@ -168,7 +171,8 @@ def update_pipeline(self):
 
         src = mm.source
         self.configure_connection(self.stream_tracer, src)
-        self.seed.inputs = [src]
+        if isinstance(self.seed, SourceWidget):
```
prabhuramachandran: Thanks for the PR!  I am not too happy about this `isinstance` check, it suggests that there should be a cleaner way to do this.  Is the intent to just allow a user to set the polydata and not use a widget at all?  I think it would be very useful thing in any case to allow users complete control over the points.  So maybe I should think of supporting either a widget or a poly data.  This could perhaps be done by simply having another attribute on the streamline module, say `user_seed` which is a polydata and if this is non-None, that we use that instead of the seed?

PR #680, opened 2018-07-12
fix upper case extension

solarjoe:
pyface will throw a KeyError if a filename with an upper case extension like "my_image.PNG" it entered in the textbox as the extension is not in the `meth_map`.


prabhuramachandran:
Thank you for the fix.

PR #708, opened 2018-09-01
DOC: Reword docs for animator function

hamogu:
The first line was too long and broken into a second line, which causes a bad rendering in sphinx:
```
:: A convenient decorator to animate a generator that performs an

    animation.
```
The rewording saves a few characters so that is all fits on one line and shows in sphinx correctly.

[skip ci]


prabhuramachandran:
Thanks!

Issue #696, opened 2018-08-03
Can't update plot3d animation


jmichaux:
I'm trying to create an animation using plot3d by modifying [this example](http://zulko.github.io/blog/2014/11/29/data-animations-with-python-and-moviepy/). 

```
import numpy as np
import mayavi.mlab as mlab

duration = 5
fig = mlab.figure(size=(500, 500), bgcolor=(1,1,1))

u = np.linspace(0,2*np.pi,100)
xx,yy,zz = np.cos(u), np.sin(3*u), np.sin(u) # Points
l = mlab.plot3d(xx,yy,zz, representation="wireframe", tube_sides=5, line_width=.5, tube_radius=0.2, figure=fig)


@mlab.animate()
def make_frame():
	""" Generates and returns the frame for time t. """
	t=0
	while True:
		y = np.sin(3*u)*(0.2+0.5*np.cos(2*np.pi*t/duration))
		# l.mlab_source.set(y = y)
		l.mlab_source.reset(x=xx,y=y,z=zz)
		t += 1
		yield

make_frame()
mlab.show()

```

I've tried using set and reset ([#592](https://github.com/enthought/mayavi/issues/592)) to update the animation but I keep getting the same error:  

`AttributeError: 'MLineSource' object has no attribute 'vectors'`

I'm fairly certain that the above syntax has worked before.  For example, [this code](https://github.com/jnez71/CartPole/blob/b65fae01ac83ad329f7503d59f33199060ea56d7/cartpole/simulator.py) was used to make [this visualization](https://imgur.com/a/9QWn3).

prabhuramachandran:
Thanks for reporting this and the script.  Looks like a bug!  I'll push a fix to master shortly.  If you are unable to run using mayavi master you can fix it by editing the `mayavi/tools/sources.py` module on your system and edit the MLineSource class (around line 456) and add this:
```
    vectors = ArrayOrNone

```

prabhuramachandran:
Note that this will fix your problem for now but this may not be the eventual fix that I push.

jmichaux:
@prabhuramachandran Thank you for looking into this. The temporary fix works nicely.

prabhuramachandran:
Fixed in #699, closing.

PR #722, opened 2018-10-09
Fix bug related to SurfaceSource.scalars

rahulporuri:
This bug manifests when the SurfaceSource object in question does not have a scalars attribute and many of them do not have this.

This bug is related to #696  and the fix is similar to #699


prabhuramachandran:
LGTM, it would be great if you could add a test case like in #699 for this.  Thanks!

PR #729, opened 2018-10-25
FIX: Fix for VTKNoneArray

larsoner:
This fixes a bug I've hit because the check for `VTKNoneArray` is in the wrong spot, so you can get an error on line 83:
```
has_nan = np.isnan(x).any()
```
That looks like:
```
>>> np.isnan(None)
TypeError: ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

### DIFF
--- a/mayavi/core/utils.py
+++ b/mayavi/core/utils.py
@@ -70,11 +70,11 @@ class DataSetHelper(object):
         if x is None:
             return None, [0.0, 1.0]
         name, x = x
-        if self._composite:
+        if isinstance(x, dsa.VTKNoneArray):
+            res = [0.0, 1.0]
+        elif self._composite:
             # Don't bother with Nans for composite data for now.
-            if isinstance(x, dsa.VTKNoneArray):
-                res = [0.0, 1.0]
-            elif attr == 'scalars':
+            if attr == 'scalars':
                 res = [algs.min(x), algs.max(x)]
             else:
                 max_norm = np.sqrt(algs.max(algs.sum(x*x, axis=1)))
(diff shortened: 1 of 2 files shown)


larsoner:
I've spend about half an hour trying to figure out how to get it to the state where there is a VTKNoneArray in there from the Mayavi tests and source with no luck. I'll take some more time trying to remember which use case elicited the error in the first place and try to do some debugging there (but IIRC, it's a pain because it's in a complex interactive Traits-based GUI).

prabhuramachandran:
Ahh, sorry to hear that.  This is why it is useful to have the test case. :)  You can create a VTK dataset perhaps to trigger it without having to go through mlab/mayavi if that helps.  I will wait a bit in case the light bulb goes on but if it doesn't, and it seems too hard, I'll see if I can think of a test and add it.  Thanks!

larsoner:
> You can create a VTK dataset perhaps to trigger it without having to go through mlab/mayavi if that helps.

I've spent another half an hour trying to do this with something like the following in `test_core_utils.py`:
```
    def test_get_range_works_for_VTKNoneArray(self):
        id_ = tvtk.ImageData(dimensions=(2, 2, 1), origin=(0, 0, 0),
                             spacing=(1, 1, 1))
        id_.point_data.scalars = dsa.VTKNoneArray()
        dsh = DataSetHelper(id_)
        self.assertEqual(dsh.get_range(), (None, [0, 1]))
```
But this gives:
```
E           TypeError: SetScalars argument 1: method requires a VTK object
```
I cannot figure out how to get the `scalars` property to be a `VTKNoneArray` object.

larsoner:
Okay I finally found a test that passes on my system. But it also exposes that there is some updating wackiness in that calling `get_range()` twice can yield different results. No idea how to fix that one, but at least this does not crash with the `ufunc 'isnan' not supported` error anymore.

prabhuramachandran:
Sorry for the delay, looks good, will merge this and look at your other PR.

prabhuramachandran:
Thanks!

PR #735, opened 2018-11-13
FIX: Fix set_range

larsoner:
In my scripts I kept seeing things like:
```
ERROR: In /work/standalone-x64-build/VTK-source/Common/Core/vtkLookupTable.cxx, line 143                                                                                            
vtkLookupTable (0x5a21890): Bad table range: [6.42653e-11, 2.01495e-11]
```
Turns out it's because internally the code somewhere does essentially:
```
self.vmin = vmin
self.vmax = vmax
```
The problem is that the first set causes a trait notification, and can lead to the condition in `_vmin_changed` that `self.vmin is not None` but `self.vmax is None`; and if the max of the data is less than `self.vmin` you end up in `lut_manager` trying to set things incorrectly.

This puts some sort of fix in both places just to be safe.

### DIFF
--- a/mayavi/core/lut_manager.py
+++ b/mayavi/core/lut_manager.py
@@ -429,6 +429,8 @@ class LUTManager(Base):
         self._default_data_range_changed(self.default_data_range)
 
     def _data_range_changed(self, value):
+        # should be guaranteed by callers, otherwise VTK will print an error
+        assert value[0] <= value[1]
         try:
             self.lut.set_range(value[0], value[1])
         except TypeError:
--- a/mayavi/tests/test_mlab_integration.py
+++ b/mayavi/tests/test_mlab_integration.py
@@ -13,6 +13,7 @@ import mock
 import numpy as np
 from numpy.testing import assert_allclose
 from traits.testing.unittest_tools import UnittestTools
+from traits.api import push_exception_handler
 
 import vtk
 
@@ -22,6 +23,8 @@ from tvtk.api import tvtk
 from mayavi.tools.engine_manager import engine_manager
 from mayavi.core.registry import registry
 from mayavi.tests.common import get_example_data
+
+push_exception_handler(reraise_exceptions=True)  # XXX should probably be elsewhere
 
 
 ################################################################################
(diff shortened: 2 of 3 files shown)


larsoner:
Pushed a test that fails if the changes from `mayavi/tools/modules.py` are omitted (it fails at the `assert value[0] <= value[1]` line)

mayavi/tests/test_mlab_integration.py
```diff
@@ -23,6 +24,8 @@
 from mayavi.core.registry import registry
 from mayavi.tests.common import get_example_data
 
+push_exception_handler(reraise_exceptions=True)  # XXX should probably be elsewhere
```
prabhuramachandran: Why do you need this?  Where do you think this should go?  Was it helpful when running the tests?

mayavi/core/lut_manager.py
```diff
@@ -429,6 +429,8 @@ def _use_default_range_changed(self, value):
         self._default_data_range_changed(self.default_data_range)
 
     def _data_range_changed(self, value):
+        # should be guaranteed by callers, otherwise VTK will print an error
+        assert value[0] <= value[1]
```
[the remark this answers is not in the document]
larsoner: It's indicative that some caller has done something bad. So really it should be fixed (as was done here). But in principle something like `value = np.array([np.mean(value)] * 2)` could be used here (probably better to make them equal rather than swap).

mayavi/tests/test_mlab_integration.py
larsoner: Ideally you could make it a suite-level fixture (`pytest`), or decorator for any file/test that uses `traits`. Otherwise you can get silent failures that only show up on `stdout` and are not actually raised.

If you remove the changes in ` mayavi/tools/modules.py`, you get a failure (TDD). If you also then uncomment this line, you just get a message on `stdout` but no failure.

prabhuramachandran: Still using nose right now, might change that at some point but for now I guess lets leave it where it is.  I'll merge it for now.  Thanks for the clarifications.

PR #747, opened 2018-12-20
FIX: Escape chars

larsoner:
Fixes errors like:
```
SyntaxError: invalid escape sequence \i
```
by using `r` strings.


prabhuramachandran:
Thanks, looks good, not sure why some of the docs are using `\in`.  The appveyor test issues seem to have to do with a change in numpy version. I will take a look later.

Issue #768, opened 2019-04-19
Can I L10N sphinx document?


tkoyama010:
Hello. Thank you for useful visualizing tool.
I want to translate mayavi's sphinx documents to my language by using [Transifex](https://www.transifex.com/).
Is there other project which is translating this project's document?
Is there any problem which I have to know to translate document?

prabhuramachandran:
Thank you for the offer to help.  I have no idea what the best way to do this is as I have not translated documentation before.  There are no other projects to my knowledge that are translating this project. Feel free to translate it.  Thank you!

tkoyama010:
Thank you for your reply. I'll make framework to translate. After that I'll contact you again.

tkoyama010:
I made a mayavi page in readthedocs which will be translated in Japanese.
https://mayavi-ja.readthedocs.io/ja/latest/index.html
Repository for translate docs for mayavi official document is here.
https://github.com/getfem-doc/mayavi-doc-translations
Minor changes are required to the conf.py file of the mayavi sphinx documentation. I will contact you again about that.

jacowan:
Dear Koyama-san

Please forgive me for jumping into your discussion with Prabhu. Your exchanged popped up in some internal Enthought threads and was forwarded to me as I coordinate recruiting at Enthought. We are currently hiring Scientific Software developers for our Tokyo office and someone with a background in Mayavi would be ideal. If you have any interest in discussing the opportunity, please let me know. I may be reached at jillc@enthought.com or jobs@enthought.com. Thank you!

PR #777, opened 2019-05-07
Update for #768

tkoyama010:
I made a translation repository for mayavi documentation.For this repository, a few update is need for conf.py in mayavi. Thank you.
https://github.com/getfem-doc/mayavi-doc-translations.git

### DIFF
--- a/docs/source/mayavi/conf.py
+++ b/docs/source/mayavi/conf.py
@@ -16,6 +16,8 @@
 # extensions.
 import os
 try:
+    # When translating the docs to another language, this variable is set 
+    # elsewhere, so if it is not set, set it to a suitable default.
     basedir
 except NameError:
     basedir = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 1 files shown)


docs/source/mayavi/conf.py
```diff
@@ -15,7 +15,10 @@
 # Adding the current directory to the path, so that sphinx finds the
 # extensions.
 import os
-
+try:
+    basedir
```
prabhuramachandran: I am curious as to why this is needed?  Can't you just define it?  If there is a non-obvious reason, then please do write a comment explaining why this is needed.

tkoyama010: Thank you for your comment. When the building of document in mayavi-doc-translations.git, basedir is set by the follwing conf.py.
https://github.com/getfem-doc/mayavi-doc-translations/blob/master/conf.py
That's the reason to check the variable.

prabhuramachandran:
Thanks for this, I will merge this as soon as the builds pass.  I've added a comment in the conf with your clarification.

Issue #791, opened 2019-06-18
Python 3 issue with str/bytes in jupyter nb


ssandfeld:
Using mayavi.mlab in a jupyter notebook with Python 3.7.3 gives "TypeError: a bytes-like object is required, not 'str'" for arbitrary mlab functions (e.g., points3d). This error occurs frequently but randomly. 
`mayavi.__version__ = 4.6.2`

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
~/.virtualenvs/science/lib/python3.7/site-packages/IPython/core/formatters.py in __call__(self, obj)
    343             method = get_real_method(obj, self.print_method)
    344             if method is not None:
--> 345                 return method()
    346             return None
    347         else:

~/.virtualenvs/science/lib/python3.7/site-packages/mayavi/tools/notebook.py in _repr_html_(self)
     57         return scene_to_png(scene)
     58     elif _backend == 'x3d':
---> 59         return scene_to_x3d(scene)
     60 
     61 

~/.virtualenvs/science/lib/python3.7/site-packages/mayavi/tools/notebook.py in scene_to_x3d(scene)
     94     else:
     95         url_base = "https://www.x3dom.org/download/1.7.2"
---> 96     x3d_elem = _fix_x3d_header(ex.output_string)
     97     html = '''
     98     %s

~/.virtualenvs/science/lib/python3.7/site-packages/mayavi/tools/notebook.py in _fix_x3d_header(x3d)
     74     x3d = x3d.replace(
     75         '<X3D profile="Immersive" version="3.0">',
---> 76         rep
     77     )    
     78     return x3d

TypeError: a bytes-like object is required, not 'str'

```

ssandfeld:
A potential fix could be to add
``` python
if isinstance(x3d, bytes):
    x3d = x3d.decode("utf-8", "ignore") 
```
in file `mayavi/tools/notebook.py`, function `_fix_x3d_header`, line 70, which for mayavi 4.6.2 seems to do the job.

PR #792, opened 2019-06-18
add fix for str conversion from bad bytes array

ssandfeld:
This fixes issue #791.


prabhuramachandran:
Thanks for the fix.  With Mayavi from master there is a more powerful notebook mode that you might like to try.  I will merge this.

PR #795, opened 2019-06-28
Test against ETS packages from source using travis cron jobs.

rahulporuri:
Install ETS packages from source when the travis event type is a cron job.


rahulporuri:
Travis CI for the commit https://github.com/enthought/mayavi/pull/795/commits/ddf9a59183d71e08f354fe108d67821e9f02d71c does what we expect - it installs the ETS packages from source and runs the tests.

PR #820, opened 2019-08-28
Modified picker 

amalss18:
Made the following changes:
1) Shifted picker GUI to configure GUI
2) Picker data appears as a text widget on the scene
3) added a mlab.set_picker_props() to set properties of the picker


mayavi/tools/helper_functions.py
```diff
@@ -1179,3 +1181,9 @@ def test_triangular_mesh():
     t = np.r_[0, t]
 
     return triangular_mesh(x, y, z, triangles, scalars=t)
+############################################################
+
+def set_picker_props(figure=gcf, pick_type="point_picker",tolerance=0.025):
+    figure().scene.picker.set_picker_props(pick_type, tolerance)
```
prabhuramachandran: Why are you passing a callable to get the figure, instead figure should be the result of a `figure()` call.  The default for `figure` could be None and in that case you can call `gcf`.

tvtk/pyface/picker.py
```diff
@@ -158,29 +159,44 @@ def handle_pick(self, data):
...
+            text_actor.set(input="ID : %s\nx : %s\ny : %s\nz : %s \
+            \nscalar : %s\nvector : %s\ntensor : %s "
+                           % (self.ID, x_coord, y_coord, z_coord,
+                              self.scalar, self.vector, self.tensor))
+        elif self.vector is not None and self.scalar is not None:
+            scalar = np.format_float_scientific(self.scalar, precision=3)
+            vector = np.zeros(3)
+            for i in range(3):
+                vector[i] = np.format_float_scientific(self.vector[i], precision=3)
+            text_actor.set(input="ID : %s\nx : %s\ny : %s\nz : %s \
```
prabhuramachandran: Same here and below.

mayavi/tools/helper_functions.py
```diff
@@ -1179,3 +1181,10 @@ def test_triangular_mesh():
     t = np.r_[0, t]
 
     return triangular_mesh(x, y, z, triangles, scalars=t)
+############################################################
```
prabhuramachandran: Please remove these unnecessary commented lines.  Also no need to add an import for gcf, can't you just use `tools.gcf`?

tvtk/pyface/picker.py
```diff
@@ -81,7 +83,7 @@ class PickHandler(HasTraits):
     what you need.  Each time a pick occurs the handle_pick is called
     by the `Picker` class."""
 
-    def handle_pick(self, data):
+    def handle_pick(self, data, renwin):
```
prabhuramachandran: I do not like that the API has changed suddenly, if anyone has implemented a custom `PickHandler`, it will fail.

prabhuramachandran: If you want to pass additional data, you can add some extra traits to the `PickedData` class and add the renwin and text actor which can be optional.

tvtk/pyface/picker.py
```diff
@@ -415,42 +432,49 @@ def pick_world(self, x, y):
...
-    def _tolerance_changed(self, val):
-        """ Trait handler for the tolerance trait."""
-        self.pointpicker.tolerance = val
-        self.cellpicker.tolerance = val
+    def text_setup(self):
+        """Sets the properties of the text widget"""
+        self.text_actor._get_text_property().font_size = 100
+        self.text_rep._get_position_coordinate().set(value=(.15, .15, 0))
+        self.text_rep._get_position2_coordinate().set(value=(.3, .2, 0))
+        self.text_widget.set(representation=self.text_rep)
```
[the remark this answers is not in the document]
prabhuramachandran: This is not resolved it seems?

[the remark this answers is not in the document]
amalss18: I've handled for this case in lines 347 - 350

PR #843, opened 2019-10-23
Avoid using eval

tkoyama010:
Hi. This is the change of avoid using eval. I changed eval to safer ast.literal_eval.
https://docs.python.org/3.8/library/ast.html#ast.literal_eval
Thank you for reading.

### DIFF
--- a/examples/mayavi/interactive/lorenz_ui.py
+++ b/examples/mayavi/interactive/lorenz_ui.py
@@ -109,7 +109,7 @@ class Lorenz(HasTraits):
             g = scipy.__dict__
             x, y, z = self.points
             s, r, b = self.s, self.r, self.b
-            val = ast.literal_eval(func_str, g,
+            val = eval(func_str, g,
                         {'x': x, 'y': y, 'z': z,
                          's':s, 'r':r, 'b': b})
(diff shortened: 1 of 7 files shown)


examples/mayavi/interactive/lorenz_ui.py
```diff
@@ -108,7 +109,7 @@ def get_vel(self, comp):
             g = scipy.__dict__
             x, y, z = self.points
             s, r, b = self.s, self.r, self.b
-            val = eval(func_str, g,
+            val = ast.literal_eval(func_str, g,
```
prabhuramachandran: This is incorrect and will not work, in cases like this `eval` is really better than `ast.literal_eval`.

tkoyama010: Thank you for your comment 👍

tkoyama010:
```suggestion
            val = eval(func_str, g,
```

tkoyama010:
Hmm... it seems CI problem.
https://travis-ci.org/enthought/mayavi/jobs/629317038#L52

PR #844, opened 2019-10-24
Button and Slider Widgets 

amalss18:
This will allow the user to add a slider, button to the scene and associate callbacks with these widgets.

### DIFF
--- a/mayavi/sources/widget_source.py
+++ b/mayavi/sources/widget_source.py
@@ -1,0 +1,51 @@
+from mayavi import mlab
+from mayavi.core.source import Source
+
+class WidgetSource(Source):
+    ##############################################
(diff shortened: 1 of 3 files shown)


mayavi/sources/widget_source.py
```diff
@@ -0,0 +1,51 @@
...
+    # Non-public interface
+    ######################################################################
+
+    def _update(self):
+        if self.scene is None:
+            return
+
+    def _scene_changed(self, old, new):
+        if self._actors_added:
+            old.remove_wdigets(self.widgets)
```
prabhuramachandran: Typo, and the fact that this wasn't caught suggests no tests, can you please add a simple test so this is exercised?

PR #849, opened 2019-11-05
:construction_worker: FIX for building latex and pdf

tkoyama010:
:construction_worker: Hello mayavi project. I fixed conf.py, figure, typo for building latex and pdf. Could you review and merge?
 Thank you for reading.


docs/source/mayavi/auto/example_superquad_with_gui.rst
```diff
@@ -8,17 +8,17 @@ Superquad with gui example
 
 This example uses MayaVi to show the evolution of a superquadric
 (http://en.wikipedia.org/wiki/Superquadrics), which are ellipsoidal surfaces parametrised
-by two parameters,lpha and eta. The equations that are used to determine the superquadric are 
+by two parameters, alpha and beta. The equations that are used to determine the superquadric are
```
prabhuramachandran: We should be able to use math symbols with suitable rst markup, we can add the `sphinx.ext.mathjax` to the sphinx extensions, and then using 
```
:math:`\alpha`
```
markup to show inline math or using separate math sections for equations and using normal LaTeX equations.

tkoyama010:
Thank you for comment. I agree that it is not good to update figure. I will make new branch and make pull request.

PR #851, opened 2019-11-13
:recycle: UPDATE code style to PEP8 by black

tkoyama010:
:recycle: Hello I updated mayavi's code style to [PEP8](https://www.python.org/dev/peps/pep-0008/) by using [black](https://github.com/psf/black). This is a change of refactoring.
I welcome comments on this change.


MuellerSeb:
The failing of TravisCI seems to be unrelated to the proposed changes.

prabhuramachandran:
Thanks for your effort on this.  Unfortunately, I am not comfortable with such a large change and also do not like the default styles with black especially on functions with many arguments, the code becomes annoyingly long.  But thats my own opinion.  I am fine with PEP8 corrections and reformatting but black is a bit too aggressive for me at this point.  

@mdickinson -- is there an enthought standard for formatting the code?

tkoyama010:
@prabhuramachandran Thank you for your comment. I agree that we should be careful about such a massive change. Once I agree on this, I will resolve the conflict.

mdickinson:
> is there an enthought standard for formatting the code?

Not really, no, other than that we usually follow PEP8 (or at least PEP8 with minor adjustments, and additional details on import ordering to cover ETS).

I did a once-in-a-decade `black` reformatting to the Traits codebase last year, but that's not something I plan to repeat, and the Traits formatting was particularly egregious (spaces between a function name and the opening parenthesis for the function call, for example, which complicates attempts to search for calls to a particular function). There are formatting choices from `black` that I dislike, but overall the change was a net improvement to the codebase, and the use of `black` in particular saved us from having detailed discussions about individual style changes (which is kinda the point).

I did try to use `autopep8` before I tried `black`, but `autopep8` introduced new `SyntaxError`s and failed in some otherwise quite bizarre ways. The fact that `black` validates that its changes don't alter the AST is a big win in my eyes; I have much more trust that `black` won't inadvertently introduce bugs than with some of the other autoformatting tools.

The biggest problem with a touch-every-file change like this is that it's likely to invalidate all open PRs, requiring merge conflicts to be resolved with those PRs, so you want to be careful about the timing of such a change. A secondary issue is that it makes tools like `git blame` harder to use.

mdickinson:
> I did a once-in-a-decade `black` reformatting to the Traits codebase last year

For reference: https://github.com/enthought/traits/pull/432

aboks:
> A secondary issue is that it makes tools like git blame harder to use.

@mdickinson FYI, the problem with `git blame` can (to a large degree) be circumvented as of git 2.23. I wrote a blog post about this a while ago: https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame.

tkoyama010:
Any update for this ?

mdickinson:
@tkoyama010: Given that @prabhuramachandran has said that he's not comfortable with this change, I'd propose that we close this.

tkoyama010:
OK. Thanks.

PR #856, opened 2019-11-23
FIX : updating position of Box

masaogata:
Small revision on Box class to apply a change of position instantly. I'm not familiar with tvtk so I referred some codes of Sphere class.

### DIFF
--- a/tvtk/tools/visual.py
+++ b/tvtk/tools/visual.py
@@ -1124,8 +1124,8 @@ class Sphere(HasTraits):
                                theta_resolution = 20)
         sp.update()
         ps = sp.output
-
         points = ps.points.to_array()
+
         self.points = points
         self.polydata.points = self.points
         self.polydata.polys = ps.polys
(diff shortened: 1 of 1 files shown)


tvtk/tools/visual.py
```diff
@@ -1392,6 +1392,7 @@ class Box(HasTraits):
     polydata = Instance(tvtk.PolyData, ())
     actor = Instance(tvtk.Actor, ()) # tvtk Actor, for the usual pipeline architecture.
     property = Instance(tvtk.Property)
+    normals = Instance(tvtk.PolyDataNormals, ())
```
prabhuramachandran: It isn't clear why adding normals should change the position at all?  I am not able to understand the commit message and change.  What is the problem you are facing and why are you making this change?  The poly data normals should not affect the box as it will only smooth small angles and not larger angles like the one for a box.

prabhuramachandran:
I've fixed the issue with Box and Helix in master.  Closing this for now.  Thank you!

masaogata:
Sorry for that I couldn't catch up this, and thank you for kind comment!

PR #858, opened 2019-12-01
:horse: avoid log file directory not exist

tkoyama010:
:horse: Hello I added line to avoid error from log file directory not exist. Thank you for reading.

### DIFF
--- a/mayavi/plugins/app.py
+++ b/mayavi/plugins/app.py
@@ -58,6 +58,8 @@ def setup_logger(logger, fname, stream=True, mode=logging.ERROR):
             return
 
     logger.setLevel(logging.DEBUG)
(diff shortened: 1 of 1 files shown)


prabhuramachandran:
I am curious as to how you ran into this error, I am surprised that no one else has run into it earlier?

tkoyama010:
Thank you for your comment. You can reproduce this error by using follwing Google Colaboratory ⬇️ 
https://colab.research.google.com/drive/1an-7YsMWTNJ4CkVUZJwo2relvehbU8QF

prabhuramachandran:
Nice collaboratory demo. Thanks. Will merge.  Sorry for the delay getting to all these.

PR #859, opened 2019-12-01
Replace deprecated rich_compare kwarg when creating traits

rahulporuri:
`rich_compare` has been deprecated since version 3.0.3 of traits and it's been replaced with the `comparison_mode` kwarg.

See PR https://github.com/enthought/traits/pull/598/ for more information.

### DIFF
--- a/mayavi/sources/array_source.py
+++ b/mayavi/sources/array_source.py
@@ -11,7 +11,7 @@ from vtk.util import vtkConstants
 
 # Enthought library imports
 from traits.api import (Instance, Trait, Str, Bool, Button, DelegatesTo, List,
-                        Int)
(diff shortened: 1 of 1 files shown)


mdickinson:
The test failure *seems* to be unrelated to the content of this PR.

I made a trivial PR #860 in an attempt to test whether the issue is with this PR in particular, or with the current CI.

mdickinson:
Okay, #860 passed, which suggests that the error here really _is_ related to the changes in the PR. I'll restart the CI, just in case this is some weird intermittent issue, but I'm not hopeful.

mdickinson:
Opened #861 for the test failure. I think this PR is fine, but the error should be investigated separately.

rahulporuri:
The CI passed! Thanks for the review @mdickinson

PR #863, opened 2019-12-03
add 'height' into cylinder_actor.

Doradx:
add optional parameter 'height' into cylinder_actor.


prabhuramachandran:
Please reset the target for master and not a release branch.  The release branch is not used.

PR #868, opened 2019-12-08
Replace uses of traits.api.Long with traits.api.Int

mdickinson:
This PR replaces uses of `Long` with `Int`. `Int`should be an acceptable substitute anywhere that `Long` is used. (On Python 2, the `Int` trait type accepts `long` instances since enthought/traits#104.)


mdickinson:
@rahulporuri @prabhuramachandran Would either of you be able to review?

PR #870, opened 2019-12-11
DOC: fix syntax error in animation example

magnunor:
Fix a syntax error in the mlab animation example, `def anim()` to `def anim():`.


prabhuramachandran:
Thanks!

PR #877, opened 2019-12-31
Fix support for wxPython 3.0

swt2c:
Retains compatibility with 2.8 - there wx.InitAllImageHandlers() is a no-op,
wx.HIDE_READONLY is 0, and wx.PySimpleApp() issues a deprecation warning.

Patch originally by Olly Betts <olly@survex.com>

### DIFF
--- a/docs/source/mayavi/auto/wx_embedding.py
+++ b/docs/source/mayavi/auto/wx_embedding.py
@@ -60,7 +60,7 @@ class MainWindow(wx.Frame):
                         kind='subpanel').control
         self.Show(True)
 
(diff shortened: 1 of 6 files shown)


prabhuramachandran:
Looks fine to me but I haven't used the wxPython backend in ages.  Does this work with wxPython 4.x?

swt2c:
These changes should be compatible with wxPython 4.x, yes.  I haven't tested it with wxPython 4.x and there are likely other changes required to make it work with 4.x.

prabhuramachandran:
Thanks, merging!

PR #879, opened 2019-12-31
WIP: Add CI build for Python 3.7

swt2c:


swt2c:
My hope was to see how the tests run with Python 3.7 but it seems there isn't an EDM runtime for Python 3.7.  :(

PR #893, opened 2020-02-10
:arrow_up: imp to importlib #892

tkoyama010:
:arrow_up: imp to importlib #892

Issue #845, opened 2019-10-25
Moving to require Python 3


tkoyama010:
Hello mayavi project. As you know, the support of python2 is end in January 1, 2020.
https://python3statement.org/
Support for numpy and scipy will also be discontinued. So that the support is end, I think we should remove the support of python2 in this project. How do you think?
Thank you for reading.

PR #894, opened 2020-02-10
:wastebasket: Moving to require Python 3 #845

tkoyama010:
:wastebasket: Moving to require Python 3 #845 . It will also solve a problem of #893 .


tkoyama010:
I can fix the problem of #893 by using six. So I close this PR. Thank you for reading.

PR #913, opened 2020-04-30
Add tests to test if hand-crafted views work.

prabhuramachandran:
This should address the lack of tests in #893.  Also cleaned up the base
module a little in the process.

### DIFF
--- a/mayavi/core/base.py
+++ b/mayavi/core/base.py
@@ -14,7 +14,7 @@ import importlib
 
 # Enthought library imports.
 from traits.api import (Any, Instance, Property, Bool, Str, Python,
-    HasTraits, WeakRef, on_trait_change)
+                        HasTraits, WeakRef, on_trait_change)
 from traitsui.api import TreeNodeObject
 from tvtk.pyface.tvtk_scene import TVTKScene
 from apptools.persistence import state_pickler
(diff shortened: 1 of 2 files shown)


prabhuramachandran:
@mdickinson -- I've added a test and also fixed an issue with the previous PR, #893. I am merging this now as master is currently broken. I would appreciate it if you could take a quick look. I can address any issues in a separate PR.  Thanks!

mayavi/core/base.py
```diff
@@ -438,32 +450,33 @@ def __view_filename_default(self):
...
     def __module_view_default(self):
         """ Try to load a view for this object.
         """
         view_filename = self._view_filename
-        try:
-            result = importlib.load_module('view', open(view_filename, 'r'),
-                            view_filename, ('.py', 'U', 1))
-            view = result.view
-        except:
+        if os.path.exists(view_filename):
```
mdickinson, line 461: It looks like this is the only functional change here - is that right? (I see a lot of other changes, but at a glance they all look cosmetic; I want to be sure I'm not missing something else significant.)

prabhuramachandran: Yes indeed, this is the only non trivial change, the rest were PEP8 errors.  Thank you!

prabhuramachandran:
The tests pass, and I added a small test case to exercise this here: https://github.com/enthought/mayavi/pull/913/files#diff-1cce85505cbc24b31c9e483ed52e7158R93-R110

prabhuramachandran:
If this all looks OK, I will delete the branch.  Thanks!

prabhuramachandran:
Thanks!

PR #919, opened 2020-05-18
Avoid a few deprecation warnings from traits.

prabhuramachandran:


prabhuramachandran:
Gah, this is too early, PrefixMap and others are not yet deprecated in the latest released traits so this PR can be merged once this is a problem.

Issue #939, opened 2020-07-09
mayavi 4.7.1 doesn't seem to work with vtk 9


rahulporuri:
This seems to have been reported by multiple users in different ways so I am creating this umbrella issue to avoid duplication and confusion. Note that depending on the Python version, it is possible to avoid this issue by using vtk 8.

rahulporuri:
issues reported earlier - https://github.com/enthought/mayavi/issues/938

rahulporuri:
Note that I am able to install mayavi from source (git master branch) with vtk 9.0.1.

rahulporuri:
one more - https://github.com/enthought/mayavi/issues/939

rahulporuri:
also reported as #922

rahulporuri:
#941 seems to be the same issue.

prabhuramachandran:
This is fixed in master, we need to cut a new release and make that available on PyPI.

liuzhenqi77:
I'm using python 3.9, vtk 9 self-built from source following [this](https://docs.pyvista.org/extras/building_vtk.html), and mayavi [master branch](https://github.com/enthought/mayavi/tree/61ef7c0eebbdca474b7cfdf6082bfc17fb5aaf59).

When importing mayavi, I met a problem related to failed `from tvtk.vtk_module import VTK_MAJOR_VERSION`, maybe an interface change in vtk 9? After fixing this, everything works fine except some minor rendering glitch.

harris-mohamed:
@liuzhenqi77 How did you fix that error? I'm having the same issue

liuzhenqi77:
@harris-mohamed I just added `VTK_MAJOR_VERSION = 9` to `mayavi/tvtk/vtk_module.py` as a temporary solution, while waiting for the official fix.

Sbozzolo:
My [GitHub Action](https://github.com/Sbozzolo/kuibit/runs/3194552872?check_suite_focus=true) stopped working when I included `mayavi`, since the installation process will never finish. I temporarily fixed by pinning `vtk` to `8.1.2`, but this works only with python3.6 or python3.7, as `vtk 8.1.2` cannot be installed for later versions.

What should to make sure that mayavi works a recent version of python?

Note that I am using mayavi on my local machine with python 3.8 and 3.9 and vtk 9. The problem is that I cannot guarantee that mayavi with vtk 9 will always work, for example, I have other machines in which I cannot even install it.

rahulporuri:
[...]

PR #909, opened 2020-04-11
ENH: VTK 9 compat

larsoner:
This is a bit nightmarish to figure out, but I'm about halfway through adding exceptions for the objects. There is probably an automated way to do it, might be worth figuring that out...

Closes #908.

PR #943, opened 2020-07-13
GitHub Action to ensure we can pip install mayavi

cclauss:
A basic test to ensure that we can `pip install mayavi` from [PyPI](https://pypi.org/project/mayavi).

Output: https://github.com/cclauss/mayavi/actions

`pip install numpy 'vtk<9.0.0'`  # Must happen first #939 and Python 3.8 is not supported.

### DIFF
--- a/.github/workflows/pip_install_mayavi.yml
+++ b/.github/workflows/pip_install_mayavi.yml
@@ -9,7 +9,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]    # [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [3.7]  # [2.7, 3.5, 3.6, 3.7, 3.8, pypy3]
+        python-version: [3.8]  # [2.7, 3.5, 3.6, 3.7, 3.8, pypy3]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
@@ -17,5 +17,5 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - run: pip install --upgrade pip setuptools wheel
-      - run: pip install numpy 'vtk<9.0.0'  # enthought/mayavi#939
+      - run: pip install numpy vtk
       - run: pip install mayavi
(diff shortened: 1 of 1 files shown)


larsoner:
> Python 3.8 is not supported.

VTK9 and 3.8 should be supported if you use latest master (or really anything after #909)

prabhuramachandran:
I am not sure I am making sense here but is it worthwhile doing `pip install https://github.com/enthought/mayavi/zipball/master` instead of just `pip install mayavi` to test the state of master?  How does installation of Mayavi from PyPI help development here?  Or is the idea to notify the developers that the PyPI release is busted and we need to update the PyPI release?  Please help me understand the intention of this PR.  Thanks!

prabhuramachandran:
And thank you very much, @cclauss,  for your contribution!

cclauss:
The idea to notify the developers that the PyPI release is busted and we need to update the PyPI release.

prabhuramachandran:
Thanks, is there a reason this is only deployed on ubuntu and not on the other platforms?  Also is worthwhile trying for different Python versions?

cclauss:
Yes.  You can uncomment lines 11 and 12 to witness broader carnage.

larsoner:
> The idea to notify the developers that the PyPI release is busted and we need to update the PyPI release.

This can be useful but it seems like it would be better to add a target that ran, say, weekly for this, rather than just running on every PR. That way if there is a "PR lull" it'll still be clear to the maintainers that it has happened, assuming they can be notified.

I know that it's possible to do this on Travis (weekly with maintainer notification of failure), not sure about GitHub actions...

For `on` PRs and master, it makes more sense to just ensure something along the lines of `pip install .` works (i.e., test the actual PR's code).

cclauss:
> pip install .

Does not solve the problem that you quoted.

Also,
* https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events
* https://docs.github.com/en/actions/getting-started-with-github-actions/about-github-actions#notifications-for-workflow-runs

larsoner:
> Does not solve the problem that you quoted.

I'm not quite sure what you mean?

I'm suggesting that we have two triggers: one weekly and one for PRs. The weekly one should `pip install` from PyPi and the PR/master-triggered one should `pip install` the code (in the PR or in `master`). Make sense?

PR #963, opened 2020-09-19
✏️ fix of grammer

tkoyama010:
Small fix of grammer


mdickinson (approved):
LGTM

mdickinson:
Assigning to myself so that I don't forget it; I'll make a release of Traits 6.3.1 tomorrow, and we should then be able to re-run the CI and get a pass.

PR #964, opened 2020-09-20
✏️ Add "Getting Started as a Translator"

tkoyama010:
Same with https://github.com/sphinx-doc/sphinx/pull/8223 .


mdickinson:
I'm having some trouble getting a clean Windows CI run. I'll give it a couple more attempts, and then push this PR through without the CI sign-off on Windows if necessary: it's clear that there's minimal risk of this PR actually being the cause of any breakage on Windows.

PR #983, opened 2020-12-10
MAINT: Support Python 3.9

larsoner:
`base64.decodestring` does not exist in Python 3.9 so current `master` causes an error on import. `decodestring` has been deprecated since 3.1 so should be safe just to use `decodebytes` directly (and same with `encodebytes`).

### DIFF
--- a/mayavi/tools/remote/ipy_remote.py
+++ b/mayavi/tools/remote/ipy_remote.py
@@ -8,12 +8,10 @@ from .remote_scene import SceneManager
 from .remote_widget import RemoteWidget
 from ..figure import gcf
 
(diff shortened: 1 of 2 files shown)


rahulporuri (commented):
I think we can simply remove the temporary variables now and just use `base64.decodebytes` or `base64.encodebytes`. Can you please make that change?

larsoner:
@rahulporuri done

rahulporuri (approved):
LGTM. Thanks @larsoner ! Waiting for CI.

prabhuramachandran:
@larsoner  -- thanks! @rahulporuri -- thanks for merging.

Issue #988, opened 2020-12-15
Test failure in cron job related to rescaling a CTF


kitchoi:
There is one test failure in the cron job built 4 days ago:

For VTK 8 build failed with this:
```
======================================================================
FAIL: Test rescaling a CTF.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/enthought/mayavi/tvtk/tests/test_ctf_util.py", line 69, in test_rescale_ctf
    self.assertEqual(edata, data)
AssertionError: {'range': (0.0, 0.75), 'rgb': [[0.0, 0.0, 0.0, 1.0], [0.2[120 chars].2]]} != {'range': (0.0, 1.0), 'rgb': [[0.0, 0.0, 0.0, 1.0], [0.25[119 chars].2]]}
  {'alpha': [[0.0, 0.0], [1.0, 0.2]],
-  'range': (0.0, 0.75),
?                  ---
+  'range': (0.0, 1.0),
?                 ++
   'rgb': [[0.0, 0.0, 0.0, 1.0],
           [0.25, 0.25, 0.5, 0.75],
           [0.5, 0.5, 1.0, 0.5],
           [0.75, 0.75, 0.5, 0.25],
           [1.0, 1.0, 0.0, 0.0]]}
```

The VTK 7 build is passing.

See https://travis-ci.com/github/enthought/mayavi/builds/208212406
See log file for the failed job: 
[mayavi-log.txt](https://github.com/enthought/mayavi/files/5696962/mayavi-log.txt)

The cron build before this one passed, and the corresponding successful log is here: 
[mayavi-log-success.txt](https://github.com/enthought/mayavi/files/5696969/mayavi-log-success.txt)

kitchoi:
Despite the job matrix `VTK` variable suggests the failing job was targeting VTK 8, the log suggested that VTK 9 was installed.

kitchoi:
While the cron job was run with Ubuntu 16, I am running into orthogonal issues with my Ubuntu 16 VM. Instead I tested this with Ubuntu 18 VM, and I could not reproduce either.

kitchoi:
Unable to reproduce on Travis CI either with the same commit and latest packages. Closing. 
If this test failure happens again in similar fashion, it might be a flaky test and there should be a separate issue.

PR #989, opened 2020-12-17
[Do-not-merge] Attempt to verify cron job failure on Travis

kitchoi:
Being unable to reproduce #988 locally, this PR intends to verify the failure on Travis. Will be closed after CI has run.


kitchoi:
Closing. Confirmed the CI build has run with comparable packages to the failed cron job and the test failure could not be reproduced.

PR #999, opened 2020-12-30
[WIP] Mayavi application workbench to tasks

aaronayres35:


aaronayres35:
I'm unsure why I'm seeing this error when I try to drag dock panes around since they theoretically have a default value of True for their `movable` trait.
```
  File "/Users/aayres/Desktop/pyface/pyface/ui/qt4/tasks/dock_pane.py", line 184, in _receive_dock_area
    self.dock_area = INVERSE_AREA_MAP[int(area)]
KeyError: 0
Abort trap: 6
```

mayavi/action/filters.py
```diff
@@ -50,6 +45,7 @@ def perform(self, event):
         mv.add_filter(obj)
         mv.engine.current_selection = obj
 
+    @on_trait_change("mayavi:engine.[current_selection,current_object]")
```
aaronayres35, line 48: This change is what is causing https://github.com/enthought/mayavi/pull/999#issuecomment-754935361

The change was motivated due to a timing issue with trying to access `self.mayavi`.  Without the change, the app can not even be launched and we have an `AttributeError: 'NoneType' object has no attribute 'on_trait_change'`.

aaronayres35: This Action is being instantiated before the GUI is started.  Thus the handler `_on_application_gui_started` in `app.py` has not yet been called, so `script.engine` is not yet set.  

We need that link up to happen earlier (ie before these actions are instantiated), or we need this notification listener to get set up later, after that sync up happens (ie after the gui has started).

mayavi/core/ui/engine_dock_pane.py
```diff
@@ -0,0 +1,33 @@
...
+    name = 'Mayavi Engine'
+
+    def _model_default(self):
+        from mayavi.core.ui.engine_view import EngineView
+        return EngineView(engine=self._get_engine(self.task.window))
+
+    def _get_engine(self, window):
+        """Return the Mayavi engine of the particular window."""
+        from mayavi.core.engine import Engine
+        return window.application.get_service(Engine)
```
corranwebster, line 17: There is a general philosophy when building Envisage apps that going through the `application` to get something within a component is a hint that things aren't architected quite right.  At a minimum it makes testing much harder, because now you need to have an (Envisage) app that has an `Engine` service available (or mock it).  More generally it makes things harder to re-use or change, because things are more tightly linked.

The approach that works better is to have things passed down from the application when they are created.  I would re-engineer this class to have an `engine` trait (or maybe even make the `engine` the `model`) and then when you create the dockpane here: https://github.com/enthought/mayavi/blob/622dc0542dc5c090ab0b08c1f3c670c8d383c88e/mayavi/plugins/mayavi_ui_tasks_plugin.py#L203
instead do the get service there (or even better, have the engine in turn be state of the `Task`, and pass the engine in to the task factory from the plugin).

mayavi/plugins/envisage_engine.py
```diff
@@ -31,7 +31,7 @@ class EnvisageEngine(Engine):
     __version__ = 0
 
     # The envisage application.
-    window = Instance(WorkbenchWindow)
+    application = Instance(TasksApplication)
```
corranwebster, line 34: I think that you probably want the Envisage `TaskWindow` here rather than the `TasksApplication`, going by straight analogy.  The reason is that a `TasksApplication` can have multiple `TaskWindows` and so some of the events you are listening to can fire when you don't expect them to.

Note that there is a minor bug here in that the old code: the `window` object is assuming an Envisage `WorkbenchWindow` and not a Pyface one (eg. `get_service` does not exist on the Pyface version).

mayavi/plugins/envisage_engine.py
```diff
@@ -118,19 +117,19 @@ def _scene_editors_changed(self, list_event):
...
-    @on_trait_change('window:opened')
+    @on_trait_change('application:window_opened')
     def _on_window_opened(self, obj, trait_name, old, new):
         """We start the engine when the window is opened."""
-        if trait_name == 'opened':
+        if trait_name == 'window_opened':
             self.start()
 
-    @on_trait_change('window:closed')
+    @on_trait_change('application:window_closed')
```
corranwebster, line 126: This will fire when _any_ window in the Tasks application closes, not just the currently used window, so you could end up in a situation where the window using this engine is still active, but the engine is not.

mayavi/plugins/envisage_engine.py
corranwebster: Addendum: or if this `EnvisageEngine` is global state for the app rather than a single window, then you need to adjust the event you are listening to for to be things like "application stop", etc. instead of "window close".

corranwebster:
A couple of general thoughts about this:

- one difference that you need to deal with is that Workbench has a UI design where you have one window per application, where tasks permits multiple task windows in an application (and potentially not ones that you control, if it comes from someone else's plugin!); this means you need to work out what stuff really belongs to the application, what stuff belongs to the task window, and what stuff belongs to the task.
- Tasks and Workbench are not necessarily mutually exclusive - this PR doesn't have to remove the workbench version - and it may be easier to create a parallel Tasks version of the application rather than trying to adapt the existing Workbench app to a Tasks app.

aaronayres35:
@corranwebster Thank you very much for the review!  I had one quick question:

> * one difference that you need to deal with is that Workbench has a UI design where you have one window per application, where tasks permits multiple task windows in an application (and potentially not ones that you control, if it comes from someone else's plugin!); this means you need to work out what stuff really belongs to the application, what stuff belongs to the task window, and what stuff belongs to the task.

So for workbench apps, there is the idea of "per-window service offers", but since there is one window per application I am a bit confused on what the original intention of this idea was? / How it could be used to achieve something different than just working with Service Offers through the Core Plugin.  My only thought was this Extension point set up for Service Offers in the WorkbenchPlugin was introduced before the one used by the CorePlugin?  This doesn't really affect this PR, but I am curious nonetheless.

aaronayres35:
> * Tasks and Workbench are not necessarily mutually exclusive - this PR doesn't have to remove the workbench version - and it may be easier to create a parallel Tasks version of the application rather than trying to adapt the existing Workbench app to a Tasks app.

This makes sense.  For now since this is a currently only a draft PR and I don't intend for it to be merged I will not readd the Workbench App here.  However, in the future when these changes are pushed through in a separate or multiple separate (cleaner) PRs, I will leave the Workbench App alongside the tasks app

aaronayres35:
Note that the TextEditor Plugin from envisage is highly workbench dependent right now, and is not included in the Tasks application in this PR.  
In theory it would make sense to have a tasks analogue of this plugin exists in `envisage`.  Likewise it might make sense to have a tasks analog of the logger from `apptools` as well.  Although that was easy enough to monkey patch to work with tasks which I tried to do in this PR.

aaronayres35:
Note that at this point: https://github.com/enthought/mayavi/pull/999/commits/622dc0542dc5c090ab0b08c1f3c670c8d383c88e

The application was in a working state.  However as mentioned in the review comments above there were multiple architecture issues, and some wrong assumptions being made that could cause problems.  In the commits since then, some refactoring has been done to clean things up but some functionality is now broken.

The current big issue has to do with the Actions contributed by the SceneUIPlugin needing access to the MayaviTask instance.  I have tried method to lazily load them, but the timing of things is still not correct.  Currently apploication can be launched but these menu items (see the Tool tab of the main menu bar) are currently disabled.

tvtk/plugins/scene/ui/actions.py
```diff
@@ -24,15 +28,14 @@ class SceneAction(Action):
...
     def _get_scene_manager(self):
         """ Trait property getter. """
-
+        #return self.task.scene_manager
         from tvtk.plugins.scene.i_scene_manager import (
             ISceneManager
         )
-
-        return self.window.get_service(ISceneManager)
+        return self.application.get_service(ISceneManager)
```
aaronayres35, line 35: I don't like this architecture / needing the application as it makes testing these actions more of a headache like Corran has mentioned, but previously these actions needed a WorkbenchWindow.  
The actions are set up to need either the application, the task or the task window (all of which are linked and can be obtained from each other).

aaronayres35:
Closing this for now as I am no longer putting time into this.  Can be reopened/used as a reference when we transition to doing mayavi work

PR #1020, opened 2021-03-24
Fix volume pipeline range check to work with cell_data

topikuu:
Previously trying to render volume views of UnstructuredGrids with
cell_data as the rendered scalar field caused constant error
messages although there are no real issues with them.

Signed-off-by: Topi Kuutela <topi.kuutela@aalto.fi>

### DIFF
--- a/mayavi/modules/volume.py
+++ b/mayavi/modules/volume.py
@@ -455,10 +455,15 @@ class Volume(Module):
         # Set the current range.
         dataset = mm.source.get_output_dataset()
         dsh = DataSetHelper(dataset)
-        name, rng = dsh.get_range('scalars', 'point')
-        if name is None:
+        point_name, point_rng = dsh.get_range('scalars', 'point')
+        cell_name, cell_rng = dsh.get_range('scalars', 'cell')
+        if point_name is None and cell_name is None:
             error('No scalars in input data!')
             rng = (0, 255)
+        elif point_name is not None:
+            rng = point_rng
+        else:
+            rng = cell_rng
 
         if self.current_range != rng:
             self.current_range = rng


rahulporuri:
@topikuu thanks for the fix but can you give us some more information about what the actual issue is? If possible, maybe add a code snippet which we can use to reproduce the issue.

topikuu:
Here's a minimal example which triggers the error:

```
from mayavi import mlab
from tvtk.api import tvtk
import numpy as np

v = np.array([[0.25, -0.5, 0.25],
              [-0.25, 0.0, 0.25],
              [0.25, 0.5, 0.25],
              [0.0, 0.0, -0.25],
              [0.25, 0.0, 0.25]], dtype=float)
t = np.array([[0, 1, 3, 4],
              [1, 2, 3, 4]], dtype=int)

t_m = tvtk.UnstructuredGrid(points = v)
t_m.set_cells(tvtk.Tetra().cell_type, t)

t_m.cell_data.scalars = np.array([1, 10])
t_m.cell_data.scalars.name = "values"
dset = mlab.pipeline.add_dataset(t_m)
volume_pipe = mlab.pipeline.volume(dset, vmin=-50, vmax=10)
volume_pipe.volume_mapper_type = 'UnstructuredGridVolumeRayCastMapper'
mlab.show()
```

The repeated messages then appear if `volume_pipe` is updated via `update_pipeline()` or `update_data()`.

rahulporuri:
@topikuu the CI failure is unrelated so you can ignore them.

mayavi/modules/volume.py
```diff
@@ -455,10 +455,15 @@ def _setup_current_range(self):
...
         dataset = mm.source.get_output_dataset()
         dsh = DataSetHelper(dataset)
-        name, rng = dsh.get_range('scalars', 'point')
-        if name is None:
+        point_name, point_rng = dsh.get_range('scalars', 'point')
+        cell_name, cell_rng = dsh.get_range('scalars', 'cell')
+        if point_name is None and cell_name is None:
             error('No scalars in input data!')
             rng = (0, 255)
+        elif not point_name is None:
```
prabhuramachandran, line 463: This should be `elif point_name is not None:`

prabhuramachandran (changes requested):
Thank you for the fix, this looks good barring the small change I have requested.  Please update this and I can merge this before the upcoming release.

mayavi/modules/volume.py
topikuu: Fixed.

rahulporuri:
thanks @topikuu

prabhuramachandran (approved):
Thanks!

PR #1022, opened 2021-04-02
Update deprecated import "collections.Callable"

rahulporuri:
Use of `collections.Callable` was deprecated in Python 3.3 and is finally removed in the upcoming Python 3.10. It needs to be imported from `collections.abc`.


aaronayres35 (approved):
LGTM

do we want this in the bug fix?

rahulporuri:
> do we want this in the bug fix?

I don't think so. Python 3.10 isn't released yet and we can work on adding Python 3.10 support for ETS later.

PR #1037, opened 2021-05-07
Bug Fix Release 4.7.3

aaronayres35:
This PR follows the lead of #960 

I suspect there is a way to auto-update change log given the formatting?  I was lazy and did not figure out if this was the case and just did it manually for the few PRs going into this release.  I also have not updated the auto generated docs :/ see second commit on linked PR.
It looks like we also don't keep `maint/<release number>` branches in mayavi as we do in other ets projects (ref: latest traitsui bugfix issue: https://github.com/enthought/traitsui/issues/1419)

In any case, this is a very quick/minor bug fix release, and I tried to follow the previous PR made in mayavi.  

I will be sure to not squash merge, and then will tag the first commit of this PR as the release.  Perhaps we need a follow up gh-pages PR? No documentation has been changed, but so that published docs have updated changelog?

### DIFF
--- a/docs/CHANGES.txt
+++ b/docs/CHANGES.txt
@@ -1,3 +1,22 @@
+Mayavi 4.7.3
+============
+
+This is a small bug fix release mainly concerned with compatability with Traits
+6.2.0 and soon to be release TraitsUI 7.2.0.
+
+Fixes
+-----
+
+10 Dec 2021 `#983 <https://github.com/enthought/mayavi/pull/983>`_ (larsoner)
+   - MAINT: Support Python 3.9
+
+06 May 2021 `#1030 <https://github.com/enthought/mayavi/pull/1030>`_ (aaronayres35)
+   - import from pyface.image not traitsui.image
+
+06 May 2021 `#1035 <https://github.com/enthought/mayavi/pull/1035>`_ (aaronayres35)
+   - Pass a value in appropriate range to avoid failure
+
+
 Mayavi 4.7.2
 ============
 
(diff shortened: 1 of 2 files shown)


rahulporuri (approved):
LGTM.

Looks like this shouldn't be squash-merged and the first of the two commits on this branch will be tagged as the release commit after the release is made.

rahulporuri:
>  Perhaps we need a follow up gh-pages PR? No documentation has been changed, but so that published docs have updated changelog?

Even if nothing has changed, it will still be good to update the gh-pages branch. Otherwise, PyPI will show that the latest release is 4.7.3 but the docs will say 4.6.2

aaronayres35:
> Even if nothing has changed, it will still be good to update the gh-pages branch. Otherwise, PyPI will show that the latest release is 4.7.3 but the docs will say 4.6.2

Yeah makes sense, I will update gh-pages in a follow up PR

prabhuramachandran:
Hi all sorry I missed this party.  It turns out that this May is 20 years since I first released Mayavi-1.0 on sourceforge.  Thanks for pushing out the release.

@aaronayres35 -- I have a script that semi- auto-generates the changelog but I think that script needs maintenance too.  Once I generate the changelog I hand edit it to retain the most useful entries.

rahulporuri:
> Hi all sorry I missed this party. It turns out that this May is 20 years since I first released Mayavi-1.0 on sourceforge. Thanks for pushing out the release.

Sorry for not giving you a headsup before pushing the release out.

20 years is a long time for an open source project. Kudos on building and maintaining this community @prabhuramachandran !

PR #1038, opened 2021-05-10
update autogenerated docs for changelog

aaronayres35:
This was missed in #1037 

It corresponds to what was the second commit on #960 (note in either case, this was done after the commit to be tagged as the release)


rahulporuri (approved):
LGTM

PR #1040, opened 2021-05-10
Use mock through unittest

aaronayres35:
`mock` is now in the standard library since python 3.3.  This PR updates the codebase to use mock via unittest.

Python 2 is end of life and we plan to drop support at some point.

This PR may be jumping the gun a bit in which case we can hold off on merging or close and re open later.


rahulporuri (approved):
Mostly LGTM with a few style-related changes. I checked out this branch and saw that there aren't any more uses of the external `mock` library

rahulporuri (approved):
still LGTM

Issue #987, opened 2020-12-14
Remove Python 2 compatibility from codebase.


rahulporuri:
Now that Python 2 is no longer supported, we can drop support for Python 2 from the codebase and remove compatibility code (e.g. uses of six. sys.version_info conditionals etc).

See related #984

rahulporuri:
closing as the package has no python 2 compatibility code anymore.

PR #1043, opened 2021-05-17
Remove use of six

aaronayres35:
part of #987 

This PR simply removes the couple uses of `six` in the codebase.

Note that `importlib.reload` is available for python 3.4+.  Thus this change assumes we are going to go ahead and drop support for python < 3.6 as we have done across the rest of ETS as python 3.5 is EOL.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,4 +7,3 @@ Sphinx
 traits>=4.6.0
 traitsui>=6.0.0
 vtk
-six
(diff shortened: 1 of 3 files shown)


rahulporuri (approved):
LGTM

rahulporuri:
looks like there are a bunch of uses of `sys.version_info` in this package which we will want to update.

PR #1045, opened 2021-05-18
Remove sys.version_info checks

aaronayres35:
part of #987 

This PR removes conditional blocks specific to older, no longer supported, versions of python

### DIFF
--- a/tvtk/tests/test_tvtk.py
+++ b/tvtk/tests/test_tvtk.py
@@ -44,10 +44,6 @@ To generate tvtk_classes.zip you must do the following::
 # Only used for testing.
 from tvtk.tvtk_classes import tvtk_helper
 
-PY_VER = sys.version_info[0]
-if PY_VER > 2:
-    long = int
-
 
 def mysum(arr):
     val = arr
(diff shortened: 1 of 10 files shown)


tvtk/tests/test_tvtk.py
```diff
@@ -337,19 +330,14 @@ def test_property(self):
         self.assertEqual(p.color, (0.5, 0.5, 0.5))
 
     def test_points_lookup(self):
-        """ Test if points can be looked up with both int and long keys.
-            Fixes GH Issue 173.
+        """ Test if points can be looked up.
```
rahulporuri, line 333:
```suggestion
        """ Test if points can be looked up. Fixes GH Issue 173.
```

rahulporuri (approved):
LGTM

tvtk/tests/test_tvtk.py
aaronayres35: I actually removed the mention of 173 on purpose as the test is now not really related to the issue:
https://github.com/enthought/mayavi/issues/173

ie the test has nothing to do with `long` anymore.

I am happy to leave the comment though if you think that makes more sense

rahulporuri: ahh. thanks for the context. resolving.

PR #1076, opened 2021-10-02
Setup GitHub Actions CI

rahulporuri:
Setup GitHub Actions CI to run the mayavi testsuite.

### DIFF
--- a/.github/workflows/run-mayavi-tests.yml
+++ b/.github/workflows/run-mayavi-tests.yml
@@ -1,0 +1,43 @@
+name: Tests
+
+on:
+  pull_request
+
+jobs:
+  tests:
+    strategy:
+      matrix: [ubuntu-latest, windows-latest, macos-latest]
+      python-version: [3.6]
+    
+    runs-on: ${{ matrix.os }}
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install Linux packages for Qt5 support
+      run: |
+        sudo apt-get update
+        sudo apt-get install qt5-default
+        sudo apt-get install libxkbcommon-x11-0
+        sudo apt-get install libxcb-icccm4
+        sudo apt-get install libxcb-image0
+        sudo apt-get install libxcb-keysyms1
+        sudo apt-get install libxcb-randr0
+        sudo apt-get install libxcb-render-util0
+        sudo apt-get install libxcb-xinerama0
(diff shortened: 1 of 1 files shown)


mayavi/tests/common.py
```diff
@@ -21,6 +30,4 @@ def get_example_data(fname):
     """Given a relative path to data inside the examples directory,
     obtains the full path to the file.
     """
-    p = os.path.join('data', fname)
-    return os.path.abspath(fixpath(p))
-
+    return os.fspath(files("mayavi.tests") / "data" / fname)
```
rahulporuri, line 33: The motivating force behind this change is the failure seen here - https://github.com/enthought/mayavi/runs/3776106273?check_suite_focus=true

```
   ...
   File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\site-packages\mayavi\core\file_data_source.py", line 199, in __set_pure_state__
    raise IOError(msg)
OSError: Could not find file at D:\C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\site-packages\mayavi\tests\data\cube.vti
```

Note the `"D:\"` at the beginning of the file path which could not be found - but the package/environment is in `"C:\"`.

rahulporuri:
Question for reviewer - given the fact that both AppVeyor and Travis CI failed on this PR (like in https://github.com/enthought/mayavi/pull/1078 and all but one jobs failed in #1080 ), i'm inclined towards removing Travis and AppVeyor CI in this PR itself.

mayavi/tests/common.py
rahulporuri: This has temporarily been addressed but I've opened #1081 to track this issue

corranwebster:
Not the reviewer, but feel free to remove TravisCI etc. if it makes your life easier.  One CI system that we are confident about is better than 3 that are fragile.

mdickinson (approved):
LGTM. The CI logs seem clean and free of any obvious issues.

We should probably expand the set of Python versions being tested, but that can definitely wait for another PR. Having working CI in place is the first priority.

rahulporuri:
> Just to be clear, we're actually using the Python installed by the setup_python rather than the system Python on the GitHub Actions workers. And that's the right thing to do, so that we have full control over the Python environment.

Yup. Sorry for that.

PR #1080, opened 2021-10-02
Cleanup print statements

rahulporuri:
This PR cleans up print staments - it remove commented out/unnecessary print statements and it converts good/necessary print statements into function calls. Note that this PR was originally motivated by the fact that


tvtk/tests/test_array_handler.py
```diff
@@ -166,7 +164,6 @@ def test_array2vtk(self):
                        if x().dtype.name not in ('float16', 'float128')]
         for dtype in (numpy.sctypes['int'] + numpy.sctypes['uint'] +
                       float_types):
-            print(dtype)
```
rahulporuri, line 169: This function call is what motivated this PR in the first place.

rahulporuri:
It looks like there are intermittent failures in the Travis CI jobs - Travis CI failed on #1078 but the PR job seems to have passed in this PR.

Issue #1082, opened 2021-10-02
Settle on the supported versions of VTK for current MayaVi


rahulporuri:
It is not clear at the moment what versions of VTK are supported by the current (unreleased) versions of MayaVi. It would be good to settle this matter - and once decided, update the CI jobs to test against all supported versions of VTK. See https://github.com/enthought/mayavi/pull/1076 for the new GitHub Actions CI jobs that are being setup to replace the Travis and AppVeyor CI jobs.

rahulporuri:
Note that at the moment, there is code which handles VTK version 5.8, version 6.2, version < 6 (representing the old pipeline) and version > 6 (representing the new pipeline). If we can settle on the supported versions, we can get rid of a lot of code e.g. remove code which handles the old pipeline.

rahulporuri:
@prabhuramachandran do you have comments on this? Can we drop the old pipeline? What versions should we be supporting?

prabhuramachandran:
@rahulporuri -- yes we can drop the support for the old pipeline. I am not sure if any of the enthought projects rely on older VTK versions, if they don't we can happily drop the old pipeline support.

rahulporuri:
> I am not sure if any of the enthought projects rely on older VTK versions, if they don't we can happily drop the old pipeline support.

I'm reasonably certain that no active Enthought project relies on the older pipeline/versions - mainly because all active development has moved to Python 3 and IIUC, the older version/pipeline (< 7) can't be used on Python 3.

rahulporuri:
Additionally, note that VTK 8.1.x and 9.0.x are the only versions available via PyPI and Python 3 is only supported by VTK version >= 7. Given that, We can definitely remove support for VTK < 7.

PR #1086, opened 2021-10-03
Remove code that handles unsupported vtk versions (< 7)

rahulporuri:
This PR removes code that handles unsupported vtk versions i.e. any vtk version less than 7. This means that we are removing support for the old pipeline, which is what most of the changes in this PR are about.

See related #1082 .

### DIFF
--- a/mayavi/tests/test_mlab_integration.py
+++ b/mayavi/tests/test_mlab_integration.py
@@ -255,9 +255,6 @@ class TestMlabNullEngineMisc(TestMlabNullEngine):
             os.remove(tmpfname)
 
     def test_slice_unstructured_grid(self):
-        v = tvtk.Version()
-        if v.vtk_major_version < 6 and v.vtk_minor_version < 10:
-            raise unittest.SkipTest('Broken on Travis with VTK-5.8?')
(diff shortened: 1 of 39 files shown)


mayavi/tests/test_mlab_integration.py
```diff
@@ -352,16 +349,9 @@ class TestMlabPipeline(TestMlabNullEngine):
     """
 
     def setUp(self):
-        ver = tvtk.Version()
-        self.less_than_or_equal_to_vtk_5_10 = True
-        if ver.vtk_major_version >= 5 and ver.vtk_minor_version > 10:
-            self.less_than_or_equal_to_vtk_5_10 = False
-        if self.less_than_or_equal_to_vtk_5_10:
-            super(TestMlabPipeline, self).setUp()
```
rahulporuri, line 360: I'm not sure of this change - because `ver.vtk_major_version >= 5 and ver.vtk.minor_version > 10` will always return `False` for the supported vtk versions i.e. vtk >= 7 but the changes I made are based on my understanding of how the code is written i.e. `if vtk version < 5.10 ... else ...`

PR #1050, opened 2021-06-16
MAINT: Update for latest VTK

larsoner:
Closes #1049

@opoplawski can you see if this fixes your build hang on Fedora 34? It at least "fixes" the hang for me on Windows. Plotting still doesn't work there (`mlab.test_plot3d()`) despite pyvistaqt.BackgroundPlotter working okay, so there is more to fix. But this might already be good enough progress in that it will prevent total `pip install` hangs for people using the VTK pre wheels. (They'll have a problem later when they try to actually plot, but that's at least easier to diagnose than an [empty build log that took an hour of CI time](https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=13871&view=logs&j=a017e066-62ca-5289-ad0b-8f57c84a089f&t=13529fc1-567a-5a12-e0e9-1de701345c43)).

PR #1088, opened 2021-10-03
Ignore classes to fix hanging build with vtk 9.0.3

rahulporuri:
This PR ignores two VTK classes - `vtkDataEncoder` and `vtkWebApplication` to prevent the package build/install from hanging with the latest version of vtk - 9.0.3. We should investigate why these classes are causing the build to hang but for now, we need to ensure that the build doesnt hang. See #1097

fixes https://github.com/enthought/mayavi/issues/1073


larsoner:
This looks redundant with #1050

rahulporuri:
> This looks redundant with #1050

I don't know which of these is the "right"/"better" solution. @prabhuramachandran comments?

rahulporuri:
closing given that we've gone with #1050

PR #1092, opened 2021-10-08
Quick fix for failures with Traits 6.3

mdickinson:
This PR represents a quick fix for failures of the Mayavi test suite with the recently-released Traits 6.3.0.

Note that this is only a band-aid workaround to get the CI up and running again. A better fix would be to rewrite `RevPrefixMap` to inherit directly from `TraitType` (which is designed to be subclassed), and not to either subclass `PrefixMap` (which wasn't designed to be subclassed), or to depend on internal details of `PrefixMap`.

### DIFF
--- a/tvtk/tvtk_base.py
+++ b/tvtk/tvtk_base.py
@@ -187,6 +187,7 @@ class RevPrefixMap(PrefixMap):
     def __init__(self, map, *extra_values, **kwargs):
         super().__init__(map, **kwargs)
         self._rmap = {}
+        self._map = {value: value for value in map}
         for key, value in map.items():
             self._rmap[value] = key
         for key in extra_values:


mdickinson:
Tests pass on Linux, failed on macOS and were cancelled on Windows. The test failure on macOS appears to be an intermittent failure that's unrelated to this PR:
```python
FAIL: Test if all the TVTK classes can be instantiated
  File "/Users/runner/work/mayavi/mayavi/tvtk/tests/test_tvtk.py", line 818, in test_all_instantiable
    raise AssertionError(message.format(''.join(errors)))
AssertionError: Not all classes could be instantiated:
  File "/Users/runner/work/mayavi/mayavi/tvtk/tests/test_tvtk.py", line 813, in test_all_instantiable
    tvtk_klass()
  File "tvtk_classes/euclidean_cluster_extraction.py", line 76, in __init__
    tvtk_base.TVTKBase.__init__(self, vtk.vtkEuclideanClusterExtraction, obj, update, **traits)
  File "/Users/runner/work/mayavi/mayavi/tvtk/tvtk_base.py", line 386, in __init__
    self.update_traits()
  File "/Users/runner/work/mayavi/mayavi/tvtk/tvtk_base.py", line 539, in update_traits
    setattr(self, name, val)
  File "/Users/runner/hostedtoolcache/Python/3.6.15/x64/lib/python3.6/site-packages/traits/base_trait_handler.py", line 75, in error
    object, name, self.full_info(object, name, value), value
traits.trait_errors.TraitError: The 'radius' trait of an EuclideanClusterExtraction instance must be 0.0 <= a floating point number <= 9.999999680285692e+37, but a value of 3.105036184601418e+231 <class 'float'> was specified.
```

I'll restart the test run.

mdickinson:
I've opened #1094 for the underlying issue. This PR should probably be updated to add a reference to that issue.

mdickinson:
Closing: we've incorporated this fix into Traits 6.3.1 and made a Traits release. It's possible that Mayavi's dependencies should be updated to explicitly exclude Traits 6.3.0.

PR #1098, opened 2021-10-20
Fix RevPrefixMap to inherit directly from TraitType

### LINKED ISSUE #1094: Rewrite RevPrefixMap to avoid depending on internal details of PrefixMap
The `RevPrefixMap` trait type subclasses `PrefixMap` and depends on its internals. `PrefixMap` wasn't designed for subclassing, and indeed changes in Traits 6.3 broke `RevPrefixMap` as a result. I'd recommend rewriting `RevPrefixMap` as a direct subclass of `TraitType`.

mdickinson:
This PR rewrites `RevPrefixMap` to inherit directly from `TraitType` rather than subclassing `PrefixMap`. This reduces the fragility of `RevPrefixMap` with respect to future upstream changes to `PrefixMap`.

To get to this code, I simply pasted in the Traits 6.2.0 version of `PrefixMap` under a new name `OldPrefixMap` to remove the dependence on upstream, and then refactored to collapse the `RevPrefixMap -> OldPrefixMap -> TraitType` inheritance tree to `RevPrefixMap -> TraitType`.

I've added some extra tests, to be sure that I didn't break existing behaviour while refactoring.

There are plenty of further cleanup and refactoring opportunities in `RevPrefixMap` (particularly in the `__init__` and with the duplication between `value_for` and `validate`), but those are probably best left to the next minor feature release. I'm hoping that this PR can go into the 4.7.4 bugfix release.

Fixes #1094.

### DIFF
--- a/tvtk/tvtk_base.py
+++ b/tvtk/tvtk_base.py
@@ -195,13 +195,14 @@ class RevPrefixMap(traits.TraitType):
 
         self._map = {key: key for key in self.map}
         self._rmap = {value: key for key, value in self.map.items()}
-        for key in extra_values:
-            self._rmap[key] = default_value
 
         if default_value is None:
             default_value = next(iter(self.map))
         else:
             default_value = self.value_for(default_value)
+
+        for key in extra_values:
+            self._rmap[key] = default_value
 
         super().__init__(default_value, **kwargs)
 


tvtk/tvtk_base.py
```diff
@@ -16,23 +16,10 @@
 from traits import api as traits
 from . import messenger
 
-try:
```
mdickinson, line 19: Hmm. I removed this `try / except` because I thought it wasn't needed any more.  But we currently explicitly declare only that Mayavi requires Traits >= 6.0, and `PrefixMap` and `PrefixList` were introduced in Traits 6.1. I think we should probably update the minimum Traits version required to Traits 6.1.

mdickinson: Done.

tvtk/tvtk_base.py
```diff
@@ -184,13 +184,47 @@ class RevPrefixMap(PrefixMap):
...
+                "Value must be {}, but a value {!r} was specified.".format(
+                    self.info(), value)
+            )
+
+        if value in self._map:
+            return self._map[value]
+
+        matches = [key for key in self.map if key.startswith(value)]
+        if len(matches) == 1:
+            self._map[value] = match = matches[0]
```
mdickinson, line 221: Traits 6.3.0 removed the caching from `PrefixMap`. We could do the same here. (But again, this would be a behaviour change that shouldn't be made in a bugfix release.)

PR #1101, opened 2021-10-29
Fix the test lockup on VTK-9.0.3.

### LINKED ISSUE #1096: Test suite hangs on macOS
I'm unable to run the test suite (under nose) on macOS, with the latest master of Mayavi in a venv: the test suite hangs with no output.

Steps to reproduce: here are the commands I'm issuing, in a zsh shell on macOS 10.15.7 (Intel MacBook Pro). Here `python3.9` is Python 3.9.7, installed via MacPorts.

```
python3.9 -m venv --clear ~/.venvs/mayavi
source ~/.venvs/mayavi/bin/activate
python -m pip install -e .
python -m pip install nose
python -m nose.core -v mayavi
```

Note: the `python -m pip install -e .` instruction also hangs intermittently (perhaps 1 run out of every 3), but succeeds often enough to allow me to get to the next step.

When the installation succeeds, the test run using `nose` hangs with no output other than a `RuntimeWarning` about `nose.core` being found in `sys.modules` prior to execution of `nose.core`. The hang can't be interrupted with a Ctrl-C, and `top` shows no CPU usage or memory growth in the associated process - IOW the hang resembles a deadlock of some kind. The Python process shows 7 OS-level threads; I'm not sure what's causing the extra threads in something that I'd expect to be single-threaded.

There's at least one obvious problem with my setup, which is that I know that the test suite expects some other packages to be present, and will fail without them. However, this still shouldn't be responsible for hanging the test run - we should have a better failure mode here.

I consider this sort of issue to be high priority to diagnose and fix: it makes development work much harder than it needs to be.

prabhuramachandran:
Some VTK classes are broken in this version and the tvtk.vtk_module
offers a simple way for us to exclude some of these classes when we go
over every VTK class.  This should fix #1096.

### DIFF
--- a/tvtk/vtk_module.py
+++ b/tvtk/vtk_module.py
@@ -27,7 +27,7 @@ except ImportError:
 vtk_version = vtkVersion.GetVTKVersion()
 SKIP = []
 
-if vtk_version == '9.0.3':
+if vtk_version in ['9.0.3', '9.0.2']:
     # Cause problems if used so ignore these.
     SKIP = ['vtkDataEncoder', 'vtkWebApplication']
     del vtkDataEncoder, vtkWebApplication
--- a/tvtk/vtk_module.py
+++ b/tvtk/vtk_module.py
@@ -9,7 +9,7 @@ need to be wrapped.
 """
 
 # Author: Prabhu Ramachandran <prabhu [at] aero.iitb.ac.in>
-# Copyright (c) 2007-2020,  Enthought, Inc.
+# Copyright (c) 2007-2021,  Enthought, Inc.
 # License: BSD Style.
 
 from vtk import *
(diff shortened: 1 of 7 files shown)


tvtk/vtk_module.py
```diff
@@ -22,3 +22,12 @@
...
     pass
+
+
+vtk_version = vtkVersion.GetVTKVersion()
+SKIP = []
+
+if vtk_version in ['9.0.3', '9.0.2']:
+    # Cause problems if used so ignore these.
+    SKIP = ['vtkDataEncoder', 'vtkWebApplication']
+    del vtkDataEncoder, vtkWebApplication
```
rahulporuri, line 33: What do you think about making this a dictionary, mapping vtk versions to the classes that should be skipped. That will be useful if we expect this to happen again in the future.

rahulporuri (approved):
LGTM with one minor comment.

tvtk/vtk_module.py
prabhuramachandran: Yes, we could but this sort of massive failure has not happened before, while wrapping has been a problem it hasn't been the case that just instantiating a class and calling some of its methods causes a lockup thus far.  So I am hoping this is not necessary going forward.  If it is, perhaps that would be a good time to redo this? Does that sound reasonable?  Right now I am seeing this as a one-off which hopefully does not recur.  We already have other issues to deal with with 9.1.x.

PR #1103, opened 2021-11-06
BUG: Set vertices for glyph source.

### LINKED ISSUE #1071: GeometryFilter clipping segmentation fault
Hi,
I noticed a segmentation fault when applying the GeometryFilter.
Here is an example reproducing the error which is adopted from [https://docs.enthought.com/mayavi/mayavi/auto/example_adjust_cropping_extents.html
](https://docs.enthought.com/mayavi/mayavi/auto/example_adjust_cropping_extents.html)

```python
import numpy as np
from mayavi import mlab

x, y, z, s = np.random.random((4, 500))

# Now build the visualization using mlab
fig = mlab.figure(1, bgcolor=(1, 1, 1))

# Create unconnected points
pts = mlab.pipeline.scalar_scatter(x, y, z, s)
mlab.outline(pts)

# Use a geometry_filter to filter with a bounding box
geometry_filter = mlab.pipeline.user_defined(pts, filter="GeometryFilter")
geometry_filter.filter.extent_clipping = True
# Connect our dialog to the filter
geometry_filter.filter.extent = [0.0, 1.0, 0.0, 1.0, 0.0, 0.8]

# The geometry_filter leaves hanging points, we need to add a
# CleanPolyData filter to get rid of these.
clip = mlab.pipeline.user_defined(geometry_filter, filter="CleanPolyData")

# Finally, visualize the remaining points with spheres using a glyph
# module
spheres = mlab.pipeline.glyph(clip, scale_factor=0.1)

mlab.show()
```

I use the versions: mayavi 4.7.3, vtk 9.0.1, python 3.8.11
Everything works fine for vtk < 9.

prabhuramachandran:
We were abusing the polys earlier to set the scalar_scatter cell
information but it seems safer and more correct to use the verts. This
fixes #1071.

### DIFF
--- a/mayavi/tests/test_mlab_source.py
+++ b/mayavi/tests/test_mlab_source.py
@@ -63,7 +63,6 @@ class TestMGlyphSource(unittest.TestCase):
 
     def test_geometry_filter_works(self):
         # This tests the bug #1071 and segfaults without the fix on VTK 9.x.
-        x, y, z, v, s, src = self.get_data()
         f = tvtk.GeometryFilter(extent_clipping=True)
         f.set_input_data_object(self.src.dataset)
         f.update()
--- a/mayavi/tests/test_mlab_source.py
+++ b/mayavi/tests/test_mlab_source.py
@@ -61,6 +61,13 @@ class TestMGlyphSource(unittest.TestCase):
         self.assertEqual(np.alltrue(vec == v), True)
         self.assertEqual(np.alltrue(sc == s.ravel()), True)
 
+    def test_geometry_filter_works(self):
+        # This tests the bug #1071 and segfaults without the fix on VTK 9.x.
+        x, y, z, v, s, src = self.get_data()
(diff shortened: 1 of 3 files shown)


mayavi/tests/test_mlab_source.py
```diff
@@ -61,6 +61,13 @@ def check_dataset(self):
         self.assertEqual(np.alltrue(vec == v), True)
         self.assertEqual(np.alltrue(sc == s.ravel()), True)
 
+    def test_geometry_filter_works(self):
+        # This tests the bug #1071 and segfaults without the fix on VTK 9.x.
+        x, y, z, v, s, src = self.get_data()
```
rahulporuri, line 66: it looks like none of these variables are used in this test. can we just remove them?

rahulporuri (approved):
Almost LGTM with one question about the regression test

prabhuramachandran:
Thanks, fixed!

PR #1104, opened 2021-11-10
CI: Dont fail other runs, fail on other commits

larsoner:
This makes GitHub Actions behave nicely like other CIs:

1. If one job fails, don't fail the other jobs (it's nice to get feedback from all)
2. If a new commit is pushed, kill the existing runs to save cycles

### DIFF
--- a/.github/workflows/run-mayavi-tests.yml
+++ b/.github/workflows/run-mayavi-tests.yml
@@ -1,4 +1,7 @@
 name: Tests
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.type }}
+  cancel-in-progress: true
 
 on:
   pull_request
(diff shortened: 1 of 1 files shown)


rahulporuri (approved):
LGTM. Thanks a lot for this @larsoner

.github/workflows/run-mayavi-tests.yml
```diff
@@ -1,4 +1,7 @@
 name: Tests
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.type }}
```
rahulporuri, line 3: can I ask why you chose this specifically?

larsoner: The three categories separate workflows, PRs, and push-vs-PR from each other. If you just had for example `${{ github.workflow }}`, pushing a commit to one PR would kill the runs in all others and and `main` builds (though you don't have any `main` builds currently)

rahulporuri: thank you very much! merging

PR #1105, opened 2021-11-10
Try fixing the crashes on VTK-9.1.0.

prabhuramachandran:
- The crashes seem to be related to the `vtkOpenGLAvatar` class being instantiated.  
- Do not instantiate all the VTK objects on import which the tvtk_doc module was doing unnecessarily.

### DIFF
--- a/tvtk/vtk_module.py
+++ b/tvtk/vtk_module.py
@@ -31,3 +31,10 @@ if vtk_version in ['9.0.3', '9.0.2']:
     # Cause problems if used so ignore these.
     SKIP = ['vtkDataEncoder', 'vtkWebApplication']
     del vtkDataEncoder, vtkWebApplication
+
+if vtk_version == '9.1.0':
+    SKIP = ['vtkOpenGLAvatar']
+    try:
+        del vtkOpenGLAvatar
+    except NameError:
+        pass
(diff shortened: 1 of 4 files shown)


prabhuramachandran:
Yay! This works, we should push out a release soon.

tvtk/vtk_module.py
```diff
@@ -31,3 +31,10 @@
     # Cause problems if used so ignore these.
     SKIP = ['vtkDataEncoder', 'vtkWebApplication']
     del vtkDataEncoder, vtkWebApplication
+
+if vtk_version == '9.1.0':
+    SKIP = ['vtkOpenGLAvatar']
+    try:
+        del vtkOpenGLAvatar
+    except NameError:
```
rahulporuri, line 39: i'm not sure why we're expecting an exception to be raised here.

rahulporuri (approved):
LGTM with one question

tvtk/vtk_module.py
prabhuramachandran: Well, I am not sure if there are build options that may exclude the building of these classes so I was just being extra safe.  The issue may arise if someone has a hand-built VTK.

PR #1135, opened 2022-02-23
Support VTK_LONG_LONG

yellowshippo:
This PR adds supports for VTK_LONG_LONG ~and VTK_UNSIGNED_LONG_LONG~ to tvtk/array_handler.py.

It is necessary because some files can contain arrays of these types.
In particular, VTU files exported using ParaView can have Int64 ~and UInt64~, as attached below.
[mesh.vtu.zip](https://github.com/enthought/mayavi/files/8126686/mesh.vtu.zip)

I also added some lines in tvtk/tests/test_array_handler.py and confirmed the VTU file attached was successfully loaded using the modified code.

Edit: I removed the description related to VTK_UNSIGNED_LONG_LONG, which is not necessary (at least) for my usage.

### DIFF
--- a/tvtk/array_handler.py
+++ b/tvtk/array_handler.py
@@ -225,7 +225,6 @@ def get_vtk_to_numeric_typemap():
         vtkConstants.VTK_LONG: LONG_TYPE_CODE,
         vtkConstants.VTK_UNSIGNED_LONG: ULONG_TYPE_CODE,
         vtkConstants.VTK_LONG_LONG: numpy.int64,
-        vtkConstants.VTK_UNSIGNED_LONG_LONG: numpy.uint64,
         vtkConstants.VTK_ID_TYPE: ID_TYPE_CODE,
         vtkConstants.VTK_FLOAT: numpy.float32,
         vtkConstants.VTK_DOUBLE: numpy.float64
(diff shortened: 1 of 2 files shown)


prabhuramachandran:
@yellowshippo -- thank you for the PR! Will run the tests to see if this passes.

prabhuramachandran (approved):
LGTM.

yellowshippo:
Thank you for your quick response. It seems the test failed for Windows. Since I have no Windows environment, I couldn't investigate the problem deeply. Any ideas to fix it?

prabhuramachandran:
@yellowshippo -- yes so this seems like a windows specific issue.  I too do not have a Windows environment handy right now.  One way to do this is to try a more defensive approach as done early in the same module, where we try to detect the type by instantiating a `vtkUnsignedLongLongArray` and checking its type and mapping to that.  You can try making a change and pushing here to see if that works.  @rahulporuri -- do you have access to a windows machine?

yellowshippo:
> One way to do this is to try a more defensive approach as done early in the same module, where we try to detect the type by instantiating a vtkUnsignedLongLongArray and checking its type and mapping to that. 

OK, I can give it a try, but where can I find the relevant code? I could not find anything related to vtkUnsignedLongLongArray in tvtk/array_handler.py and tvtk/tests/test_array_handler.py.

prabhuramachandran:
Sorry I mean, if you look at line 35, you will see this: `VTK_ID_TYPE_SIZE = vtk.vtkIdTypeArray().GetDataTypeSize()`, perhaps something similar, there is also a `GetDataTypeAsString()` method that may be useful?

yellowshippo:
@prabhuramachandran 
Thank you for the clarification. After some consideration, I found that I only need VTK_LONG_LONG and I have no use case for VTK_UNSIGNED_LONG_LONG, thus I simply removed lines related to VTK_UNSIGNED_LONG_LONG. Now it should work.

prabhuramachandran:
Merged, thank you!

Issue #1156, opened 2022-06-27
Mayavi fails building on Ubuntu 22 LTS


Becheler:
It seems there is a dependency conflict between:

- python 3.10.4
- numpy 1.23.0
- numpy.distutils

I struggle identifying the exact cause of the build failing.

```
Collecting mayavi
  Using cached mayavi-4.7.4.tar.gz (7.1 MB)
  Preparing metadata (setup.py) ... done
Requirement already satisfied: apptools in ./virtual-environment/lib/python3.10/site-packages (from mayavi) (5.1.0)
Requirement already satisfied: envisage in ./virtual-environment/lib/python3.10/site-packages (from mayavi) (6.0.1)
Requirement already satisfied: numpy in ./virtual-environment/lib/python3.10/site-packages (from mayavi) (1.23.0)
Requirement already satisfied: pyface>=6.1.1 in ./virtual-environment/lib/python3.10/site-packages (from mayavi) (7.4.1)
Requirement already satisfied: pygments in ./virtual-environment/lib/python3.10/site-packages (from mayavi) (2.12.0)
Requirement already satisfied: traits>=6.0.0 in ./virtual-environment/lib/python3.10/site-packages (from mayavi) (6.3.2)
Requirement already satisfied: traitsui>=7.0.0 in ./virtual-environment/lib/python3.10/site-packages (from mayavi) (7.3.1)
Requirement already satisfied: vtk in ./virtual-environment/lib/python3.10/site-packages (from mayavi) (9.2.0rc1)
Requirement already satisfied: configobj in ./virtual-environment/lib/python3.10/site-packages (from apptools->mayavi) (5.0.6)
Requirement already satisfied: setuptools in ./virtual-environment/lib/python3.10/site-packages (from envisage->mayavi) (59.6.0)
Requirement already satisfied: wslink>=1.0.4 in ./virtual-environment/lib/python3.10/site-packages (from vtk->mayavi) (1.6.5)
Requirement already satisfied: matplotlib>=2.0.0 in ./virtual-environment/lib/python3.10/site-packages (from vtk->mayavi) (3.5.2)
Requirement already satisfied: fonttools>=4.22.0 in ./virtual-environment/lib/python3.10/site-packages (from matplotlib>=2.0.0->vtk->mayavi) (4.33.3)
[...]

PR #1159, opened 2022-07-10
BUG: Temporary fix for VTK 9.2.0rc1

prabhuramachandran:
The vtkPlotBar class seems to have some problems so we ignore it for
now.


prabhuramachandran:
This should fix #1156.

prabhuramachandran:
@rahulporuri -- this allows folks to install mayavi with the latest VTK, please review. Thanks.

rahulporuri (approved):

PR #1174, opened 2022-09-28
Update protein.py

giammi56:
GzipFile reads files as objects ('rb' mode, text not implemented), therefore the .split() contains a "b" character in front. The solution proposed is decoding the byte into strings.

### DIFF
--- a/examples/mayavi/mlab/protein.py
+++ b/examples/mayavi/mlab/protein.py
@@ -73,17 +73,17 @@ last_atom_label = None
 last_chain_label = None
 for line in infile:
     line = line.split()
-    if line[0] in ('ATOM', 'HETATM'):
-        nodes[line[1]] = (line[2], line[6], line[7], line[8])
-        atoms.add(line[2])
-        chain_label = line[5]
+    if line[0].decode("utf-8") in ('ATOM', 'HETATM'):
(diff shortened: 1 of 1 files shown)


examples/mayavi/mlab/protein.py
```diff
@@ -73,17 +73,17 @@
 last_chain_label = None
 for line in infile:
     line = line.split()
```
JAGulin, line 75: This old example fails for me to, but I'd rather suggest converting the line in one go and drop the other changes. E.g.

line = line.decode("utf-8")

JAGulin (changes requested):
When updating anyway, the download URL should be changed to

https://files.rcsb.org/pub/pdb/data/structures/divided/pdb/q0/pdb%s.ent.gz

according to the documentation on

https://www.wwpdb.org/ftp/pdb-ftp-sites

examples/mayavi/mlab/protein.py
prabhuramachandran: Yes I agree, this is much easier and works.

prabhuramachandran (changes requested):
Thank you, I agree with @JAGulin that the one-line fix is much easier.

prabhuramachandran (approved):

prabhuramachandran:
I have made the necessary changes myself. Will fire the builds now.

PR #1175, opened 2022-09-28
Add ipyevents requirement

Ilyabasharov:
Hey, i should idditionally install ipyevents lib before using your lib. Also i've fased with gcc-8 is needed to install your lib, so i think it should be added to installation docs


prabhuramachandran:
These are optional requirements so we are not mandating these just yet. We could add a separate optional section for this but I do not want to make these hard dependencies.

PR #1183, opened 2022-10-14
Add a pyproject.toml.

prabhuramachandran:


prabhuramachandran:
@larsoner -- I am adding a pyproject.toml to see if it helps. Would be good if you could take a quick look.

prabhuramachandran:
@larsoner -- this is rubbish and doesn't work, so I will need to research this a bit more.

larsoner:
I thought about doing it because I recently did it for `codespell` and it wasn't too bad. But the presence of the C extension will make it harder here I think

prabhuramachandran:
OK all tests have passed now.

rahulporuri (approved):

larsoner (approved):
Looks good to me, +1 for merge

As a slight aside, my idea about it being difficult was because I thought you might be trying to go *entirely* to a pyproject-based build with a nearly-empty `setup.py`, e.g. like this:

https://github.com/codespell-project/codespell/pull/2523/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

You can see that `setup.py` ends up being a 2-line convenience wrapper, and that all its contents effectively land in `pyproject.toml`. I'm not sure how we would do this with Mayavi, but it would be cool if we could someday!

larsoner:
Agreed splitting into two packages sounds more maintainable!

rahulporuri:
@prabhuramachandran I keep comparing tvtk+mayavi to kiva+enable in my head but kiva doesn't need as much upkeep as tvtk does, from what I've seen over the past 3 years.

larsoner:
Yeah, I wouldn't be at all surprised if Mayavi code from today worked on tvtk+VTK from 5 years ago. It seems like the core of Mayavi has been pretty stable and updates have been to the Qt bindings, which for the most part so far have been separate from VTK/tvtk updates. But I haven't been paying very close attention so I might be wrong about this

prabhuramachandran:
@larsoner, @rahulporuri  -- yes I think we can clean up the setup.py and remove a bunch of cruft, we can also use the pyproject.toml along with the customization options for setuptools to build the ZIP files too. There was a time when tvtk was a separate package but it was a bit much to manage at that time (this was before git/github) so it was merged into the same package. I am fine with splitting it back. Retaining the source history will be hard but most of the early histort with mayavi is anyway hard to find so I guess it does not matter. The tvtk stuff changes only with VTK releases but the mayavi code changes when there are core mayavi feature improvements.

prabhuramachandran:
Merging this for now, I will try to cut a release tomorrow.

larsoner:
Yes and a more complicated set of tests I suppose, too, because now `mayavi` at least needs to be tested against several `tvtk` versions as well. So maybe better left as a TODO for the distant future :)

prabhuramachandran:
I have pushed a 4.8.1 release. I just realized that the release is stupidly including the tvtk_classes.zip but it is too late to fix it now so have added a ticket for it #1185 that I will look at later on -- before the next release. Anyway, 4.8.1 is now on pypi and hopefully should fix the issues people are facing. Thanks for all your help @larsoner.

Issue #1163, opened 2022-07-22
Usage as envisage plugin (e.g. Explorer example) not working anymore


JaRoSchm:
Hey,

the usage of mayavi as an envisage plugin isn't working anymore. This can be noticed for example
in examples/mayavi/explorer/ (there is an additional bug: explorer_app.py:110 has to be removed first).

The first run works without problem. If you now want to restart the app it fails:
```sh
  File "/home/jschmidt/Devel/mayavi_bug/explorer3d.py", line 128, in <module>
    main()
  File "/home/jschmidt/Devel/mayavi_bug/explorer3d.py", line 124, in main
    application.run()
  File "/home/jschmidt/anaconda/envs/CPG_stable/lib/python3.9/site-packages/envisage/ui/workbench/workbench_application.py", line 106, in run
    window.open()
  File "/home/jschmidt/anaconda/envs/CPG_stable/lib/python3.9/site-packages/pyface/workbench/workbench_window.py", line 159, in open
    self._create()
  File "/home/jschmidt/anaconda/envs/CPG_stable/lib/python3.9/site-packages/pyface/ui/qt4/application_window.py", line 135, in _create
    contents = self._create_contents(self.control)
  File "/home/jschmidt/anaconda/envs/CPG_stable/lib/python3.9/site-packages/pyface/workbench/workbench_window.py", line 244, in _create_contents
    self._restore_contents()
  File "/home/jschmidt/anaconda/envs/CPG_stable/lib/python3.9/site-packages/pyface/workbench/workbench_window.py", line 779, in _restore_contents
    self.layout.set_editor_memento(self._memento.editor_area_memento)
  File "/home/jschmidt/anaconda/envs/CPG_stable/lib/python3.9/site-packages/pyface/ui/qt4/workbench/workbench_window_layout.py", line 293, in set_editor_memento
    self._qt4_editor_area.restoreState(editor_layout, resolve_id)
  File "/home/jschmidt/anaconda/envs/CPG_stable/lib/python3.9/site-packages/pyface/ui/qt4/workbench/split_tab_widget.py", line 128, in restoreState
    assert self.count() == 0
```

The problem seems to be that the Mayavi scene is created first such that there already is an object
which is not allowed for restoring the window layout of the previous run. At least this is shown if one
adds
```python
print(self.widget(0))
print(self.widget(0).count())
print(self.widget(0).tabText(0))
```
[...]

PR #1207, opened 2023-02-03
Remove redunant variable in explorer example

JaRoSchm:
Hi,

additionally to #1163 this stopped the explorer example from working as since commit 0697654 numpy is imported as np which is used as a variable in this function too. I removed the variable because it was not used anyway.


corranwebster (approved):
Looks good - thanks for the contribution!

prabhuramachandran:
LGTM, the build failures are clearly unrelated.

prabhuramachandran:
Thank you. I am unable to find a way to rerun these tests, so I am merging this for now.

PR #1214, opened 2023-03-10
add jupyter support version

homosapien-lcy:
Update document to include an explanation of Jupyter notebook support for issue #441 #415

Issue #1209, opened 2023-02-13
Sporadic errors when building TVTK classes


yoda-vid:
I've been encountering sporadic installation errors the `Building TVTK classes` step. It occurs on at least Windows and macOS, both locally and in our CI. The error appears to resemble those I've seen in the Mayavi CI (eg https://github.com/enthought/mayavi/actions/runs/4081565144/jobs/7042041456#step:5:4492 and https://github.com/enthought/mayavi/actions/runs/4149642172/jobs/7178740011#step:5:4492).

Here is the error output on macOS 12.6, Python 3.8:

```
  Building wheel for mayavi (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Building wheel for mayavi (pyproject.toml) did not run successfully.
  │ exit code: -11
  ╰─> [37 lines of output]
      running bdist_wheel
      running build
      Building tvtk_classes.zip
      ----------------------------------------------------------------------
      Deleting possibly old TVTK classes
      Building TVTK classes... Fatal Python error: Segmentation fault
      
      Current thread 0x0000000103970600 (most recent call first):
        File "/private/var/folders/xc/yl9qgs0s77z7lvnz5q89tsd00000gn/T/pip-install-d759w36u/mayavi_77d97eab51f6484384d0a9226cd2cc70/tvtk/vtk_parser.py", line 711 in _find_get_set_methods
        File "/private/var/folders/xc/yl9qgs0s77z7lvnz5q89tsd00000gn/T/pip-install-d759w36u/mayavi_77d97eab51f6484384d0a9226cd2cc70/tvtk/vtk_parser.py", line 491 in _organize_methods
        File "/private/var/folders/xc/yl9qgs0s77z7lvnz5q89tsd00000gn/T/pip-install-d759w36u/mayavi_77d97eab51f6484384d0a9226cd2cc70/tvtk/vtk_parser.py", line 157 in parse
        File "/private/var/folders/xc/yl9qgs0s77z7lvnz5q89tsd00000gn/T/pip-install-d759w36u/mayavi_77d97eab51f6484384d0a9226cd2cc70/tvtk/wrapper_gen.py", line 341 in _gen_methods
        File "/private/var/folders/xc/yl9qgs0s77z7lvnz5q89tsd00000gn/T/pip-install-d759w36u/mayavi_77d97eab51f6484384d0a9226cd2cc70/tvtk/wrapper_gen.py", line 242 in generate_code
[...]

PR #1199, opened 2022-12-20
ENH: Add support for 3.11

larsoner:
Closes https://github.com/enthought/mayavi/issues/1188

First step, show the failure from https://github.com/conda-forge/mayavi-feedstock/pull/70

PR #1215, opened 2023-03-10
MAINT: test installations

homosapien-lcy:
The CI test fails due to a missing package (such as issue #1214 ), open this branch to test the solutions.

### DIFF
--- a/.github/workflows/run-mayavi-tests.yml
+++ b/.github/workflows/run-mayavi-tests.yml
@@ -64,7 +64,7 @@ jobs:
         python -m pip install pytest
         python -m pip install traitsui==7.2.1
     - name: Install mayavi and tvtk
-      run: python -m pip install -ve .[app]
+      run: python -um pip install -ve .[app]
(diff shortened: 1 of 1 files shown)


homosapien-lcy:
OK, now we don't have a denpendency issue with qt5-default, but comes our good old friend python segmentation fault (details in github action)... Just need to figure that out

homosapien-lcy:
Wow, I just realized it's not even the same segfault (this one comes from GetCellLengthComputationMode), interesting...

homosapien-lcy:
@mdickinson I found that mayavi actually past the test in issue #1199 (that ci test is past on Jan 13th though, sth should have happened in between)

yoda-vid:
Is #1209 related to this? I've encountered inconsistent segfaults during installation that I haven't pinned down but wonder if it's a resource/memory related issue.

prabhuramachandran:
This is already present in #1199 so closing.

Issue #1232, opened 2023-05-02
pip install mayavi fails building wheel


sean-reed:
Installing Mayavi with `pip install mayavi` gives the following error:

```
Building wheels for collected packages: mayavi
  Building wheel for mayavi (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Building wheel for mayavi (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [774 lines of output]
      <string>:13: DeprecationWarning:
      
        `numpy.distutils` is deprecated since NumPy 1.23.0, as a result
        of the deprecation of `distutils` itself. It will be removed for
        Python >= 3.12. For older Python versions it will remain present.
        It is recommended to use `setuptools < 60.0` for those Python versions.
        For more details, see:
          https://numpy.org/devdocs/reference/distutils_status_migration.html
      
      
      running bdist_wheel
      running build
      Already built tvtk_classes.zip
      running config_cc
      INFO: unifing config_cc, config, build_clib, build_ext, build commands --compiler options
      running config_fc
      INFO: unifing config_fc, config, build_clib, build_ext, build commands --fcompiler options
      running build_src
      Already built tvtk_classes.zip
      INFO: build_src
      INFO: building extension "tvtk.array_ext" sources
      INFO: building data_files sources
      INFO: build_src: building npy-pkg config files
      running build_py
      creating build
      creating build/lib.linux-x86_64-cpython-311
      creating build/lib.linux-x86_64-cpython-311/tvtk
      copying tvtk/class_tree.py -> build/lib.linux-x86_64-cpython-311/tvtk
      copying tvtk/code_gen.py -> build/lib.linux-x86_64-cpython-311/tvtk
 _[cut copying lines out of report]_
      copying mayavi/modules/ui/surface.py -> build/lib.linux-x86_64-cpython-311/mayavi/modules/ui
      copying mayavi/modules/ui/iso_surface.py -> build/lib.linux-x86_64-cpython-311/mayavi/modules/ui
      copying mayavi/modules/ui/__init__.py -> build/lib.linux-x86_64-cpython-311/mayavi/modules/ui
      running build_ext
      INFO: customize UnixCCompiler
      INFO: customize UnixCCompiler using build_ext
      INFO: CCompilerOpt.cc_test_flags[1073] : testing flags (-march=native)
[...]

PR #1233, opened 2023-05-07
Fix vtk 9.2.6 issues

prabhuramachandran:
This should fix recent build and test failures and also #1232.


prabhuramachandran:
@homosapien-lcy, @mdickinson, @rahulporuri -- I've fixed what I can at this point. I am not sure why the pyside6 ubuntu Python 3.10 test is segfaulting, it doesn't seem to be anything to do with Mayavi/VTK. I will need to take a closer look at the headless test failure, the same test passes for me on Python 3.11 with VTK-9.2.6.  Please review this PR and let me know how we should proceed. Thanks.

prabhuramachandran:
Closing this as #1199 fixes these issues already.

Ataraxiainc1:
The error you are experiencing is a compilation error during the installation of the "mayavi" library. It appears to occur in the file "tvtk\src\array_ext.c" and is related to data type conversions.

The warning messages indicate possible data loss in the conversions from "size_t" to "int" or from "Py_ssize_t" to "int". These warnings can be ignored as they are not critical errors.

However, the actual error occurs on line "tvtk\src\array_ext.c(2313)" where you encounter an error "C2198: 'PyCode_New': too few arguments for call". This suggests that the function "PyCode_New" is being called with fewer arguments than required, resulting in a compilation error.

To fix this issue, you will need to review the source code of "mayavi" and ensure that the call to the "PyCode_New" function has the correct number of arguments. It may be necessary to update the "mayavi" library or contact the developers for further assistance with this specific problem.

Remember that the warning messages are just warnings and should not affect the overall functionality of the library once it is successfully compiled.

PR #1234, opened 2023-05-08
Fix issue with `mlab.process_ui_events`

prabhuramachandran:
The method was basically doing nothing when no UI toolkit was explicitly set. Since this import is early it would so happen that the toolkit was not set at this point. As a result `mlab.process_ui_events` was doing nothing. In keeping with the Zen of Python, we now do not import pyface only when the toolkit is explicitly set to 'null' or the CI env var is set.

### DIFF
--- a/.github/workflows/headless-tests.yml
+++ b/.github/workflows/headless-tests.yml
@@ -14,12 +14,13 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: ['3.11']
+        python-version: ['3.10']
       fail-fast: false
 
     runs-on: ${{ matrix.os }}
     env:
       VTK_PARSER_VERBOSE: 'true'
+      ETS_TOOLKIT: 'null'
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


prabhuramachandran:
@mdickinson @corranwebster any ideas why the pyside6 tests on ubuntu is crashing so badly? The other required tests are passing. Thanks.

mdickinson:
For 3.11, you need [this apptools fix](https://github.com/enthought/apptools/issues/325). Looks like we could do with an apptools bugfix release.

mdickinson:
For the 3.10 segfault, I haven't seen anything that particularly resembles it elsewhere in ETS, so my guess is that it's tied in with the 3d stuff. Are you able to reproduce the segfault in a VM?

prabhuramachandran:
@mdickinson -- thank you. Sorry I did not see your response here so submitted an issue on apptools for a release.

As regards the segfault, I am on an ubuntu 22.04 system so will try recreating the exact same setup if possible later this week. The segfault happens during test collection so it might have to do with library conflicts between Qt and VTK. I am not sure.

prabhuramachandran:
I tried installing pyside6 in a python3.11 environment with the latest VTK along with the qt5 packages on the system, I set the env var `QT_API=pyside6`. With all that, `pytest -v mayavi` works for me without a segfault.

rahulporuri (approved):

prabhuramachandran:
Thanks, I will look at the crashing pyside6 issue in another PR. Will merge this for now.

PR #1266, opened 2023-07-24
BUG: Use numpy.import_array instead of import_array()

larsoner:
Struggling through https://github.com/conda-forge/mayavi-feedstock/pull/75 I hit this clang error:
```
  #warning "Using deprecated NumPy API, disable it with " \
   ^
  tvtk/src/array_ext.c:2712:3: error: incompatible pointer to integer conversion returning 'void *' from a function with result type 'int' [-Wint-conversion]
    import_array();
```
Which led me to find https://stackoverflow.com/questions/52828873/how-does-import-array-in-numpy-c-api-work and try the workarounds there, have them ultimately not work, and finally just use of `cimport numpy; numpy.import_array()` in https://github.com/conda-forge/mayavi-feedstock/pull/75 . Not 100% sure it's the right change but hopefully it's okay.

### DIFF
--- a/tvtk/src/array_ext.c
+++ b/tvtk/src/array_ext.c
@@ -741,7 +741,16 @@ static CYTHON_INLINE float __PYX_NAN() {
 #define __PYX_HAVE__array_ext
 #define __PYX_HAVE_API__array_ext
 /* Early includes */
+#include <string.h>
+#include <stdio.h>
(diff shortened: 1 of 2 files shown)


prabhuramachandran (approved):
This looks OK to me.  Thanks!

PR #1272, opened 2023-08-22
BUG: Fix bug with latest VTK

larsoner:
Found a fun little bug with latest `vtk` when installing from source


prabhuramachandran (approved):
LGTM, thanks!

PR #1273, opened 2023-08-22
ENH: Add support for PyQt6

larsoner:
Ports some changes over from https://github.com/Kitware/VTK/blob/master/Wrapping/Python/vtkmodules/qt/QVTKRenderWindowInteractor.py to get PyQt6 working. I also bumped the min to 3.8 since 3.7 is EOL and added a CI run for PyQt6 (hopefully it passes!).

### DIFF
--- a/.github/workflows/run-mayavi-tests.yml
+++ b/.github/workflows/run-mayavi-tests.yml
@@ -11,10 +11,13 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ['3.7']
+        python-version: ['3.8']
         qt-api: ['pyqt5']
         include:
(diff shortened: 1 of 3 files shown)


prabhuramachandran (approved):
Looks great, thank you so much!  Do I also need to merge #1266 and #1272, it looks like I should but I thought I would check with you.

larsoner:
Yes those are both still needed I think

prabhuramachandran:
Great, I have merged all of these. Many thanks. Do you think we should just cut a release of Mayavi as it currently is? I don't have time to make any specific changes.

larsoner:
Yes I think so! I think those PRs included the patches currently deployed in the conda-forge feedstock so some of them have been in use for some months already

PR #1281, opened 2023-10-19
MAINT: Support Python 3.12

larsoner:
This isn't a good fix, but `numpy.distutils` is gone in Python 3.12. This should at least allow it to build in non-editable mode.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,11 +12,16 @@ try:
     import numpy
     from numpy.distutils.command import build, install_data, build_src
     from numpy.distutils.core import setup
+    my_build_src_super = build_src.build_src
+    build_src_run = build_src.build_src.run
+    del build_src
     HAS_NUMPY = True
 except ImportError:
     HAS_NUMPY = False
     from distutils.command import build, install_data
     from distutils.core import setup
+    my_build_src_super = object
+    build_src_run = lambda *args, **kwargs: None
 import io
 import os
 import time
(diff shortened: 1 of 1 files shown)


prabhuramachandran:
Thanks @larsoner! What would be the best way to do this while supporting editable builds?

larsoner:
Not sure. And several variants of what I have here do not seem to work, e.g. the patches used here:

https://github.com/conda-forge/mayavi-feedstock/pull/77

I think in theory the right move here is to migrate to using `setuptools` directly instead of anything `numpy.distutils`-related. But I'm not sure how trivial that will be...

tkoyama010:
To support Python3.12. We should be noted whether vtk's 9.3 will be compatible with Python 3.12.
https://pypi.org/project/vtk

larsoner:
Probably not on PyPI but conda-forge has it already, which triggered https://github.com/conda-forge/mayavi-feedstock/pull/77. See 

- https://github.com/conda-forge/vtk-feedstock/pull/316
- https://anaconda.org/conda-forge/vtk/files

I haven't tested it locally but I would be surprised if it did not work.

prabhuramachandran (approved):
Am fine with these changes if it does the job.

prabhuramachandran:
@larsoner -- many thanks for this. Sorry for the delay in merging this.

PR #1290, opened 2024-01-26
VTK 9.3: Fix test suite failures (VTK API changes)

bnavigator:
We encounter  a few test suite errors in the rpm build for Mayavi since VTK has been updated to 9.3.0 in openSUSE Tumbleweed.

* VTK deprecated some threshold filter methods in 9.1 and removed them in 9.3
* There is a new attribute found by the parser
* Somehow the test contour now returns slightly rounded off values.

This is by far not complete in terms of deprecations and removals for 9.3 (See #1286 for example), but at least it greens the test suite for us.

PR #1291, opened 2024-02-08
MAINT: Add NumPy include dir during build

larsoner:
Needed to get https://github.com/conda-forge/mayavi-feedstock/pull/77 builds working. Not sure why it wasn't required before/elsewhere! But should be safe/more complete to always have it I think.


setup.py
```diff
@@ -270,133 +246,52 @@ def build_tvtk_classes_zip():
...
-
-        ext_file = os.path.join(
-            MY_DIR,
-            "tvtk",
-            "array_ext" + (".pyd" if sys.platform == "win32" else ".so")
-        )
-
-        if os.path.exists(ext_file):
-            print("Removing in-place array extensions {}".format(ext_file))
-            os.unlink(ext_file)
```
larsoner, line 366: This wasn't working anyway, on my system the generated file is `tvtk/array_ext.cpython-311-x86_64-linux-gnu.so` for example. We can revisit at some point if needed

larsoner:
cherry-picked 3 commits from https://github.com/enthought/mayavi/pull/1290 in an effort to get CIs green (can remove/revert later if that PR isn't merged before this one)

larsoner:
Just kidding... 3.12 (well really VTK 9.3) is still problematic. Based on some traceback and some local testing I think something is overwriting some memory where it shouldn't or something because the default trait it's trying to set are junk :scream: I vaguely recall having to fix something like this a while ago, I'll need to look up how I did it.

opoplawski:
Thank you for working on this - need this for Fedora support.  However it seems like the array_ext module isn't getting generated at all for me.  Is this intentional?

larsoner:
> However it seems like the array_ext module isn't getting generated at all for me. Is this intentional?

No that's not intentional, it should be :( Locally I get that it is generated, and on CIs you can see it for example [here](https://github.com/enthought/mayavi/actions/runs/7907603771/job/21585003852?pr=1291#step:6:8382). It should show up at the very least if you do `python setup.py build_ext` for example. On my macOS machine (though I got something similar on Linux last I looked):
```
$ python setup.py build_ext
running build_ext
building 'tvtk.array_ext' extension
creating build
creating build/temp.macosx-11.0-arm64-cpython-311
creating build/temp.macosx-11.0-arm64-cpython-311/tvtk
creating build/temp.macosx-11.0-arm64-cpython-311/tvtk/src
clang -DNDEBUG -fwrapv -O2 -Wall -fPIC -O2 -isystem /Users/larsoner/Applications/MNE-Python/1.6.0_0/.mne-python/include -arch arm64 -fPIC -O2 -isystem /Users/larsoner/Applications/MNE-Python/1.6.0_0/.mne-python/include -arch arm64 -I/opt/OpenBLAS/include -I/opt/homebrew/opt/llvm/include -I/Users/larsoner/Applications/MNE-Python/1.6.0_0/.mne-python/lib/python3.11/site-packages/numpy/core/include -I/Users/larsoner/Applications/MNE-Python/1.6.0_0/.mne-python/include/python3.11 -c tvtk/src/array_ext.c -o build/temp.macosx-11.0-arm64-cpython-311/tvtk/src/array_ext.o
...
$ ls build/lib.macosx-11.0-arm64-cpython-311/tvtk/
array_ext.cpython-311-darwin.so
```

opoplawski:
Okay, I think it's an issue with the Fedora pyproject build macros, but I can stick with the classic python build macros for now.  This fixes the Fedora build so it would be nice to see it merged.  FWIW - we are still at VTK 9.2.6.

larsoner:
I opened upstream https://gitlab.kitware.com/vtk/vtk/-/issues/19252 but I'll try to find a way to patch it here...

larsoner:
Okay:

1. Added a workaround for the vtkCylinderSource problem
2. Added the shiny new macos-14 / arm64 runner for testing purposes
 
Everything is *finally* green for real on VTK 9.2.6 and 9.3 and Python 3.12. The corresponding [mayavi-feedstock ](https://github.com/conda-forge/mayavi-feedstock/pull/77) is also green so I think we should be good here!

I know the `+7,129 −3,542` is scary but this is mostly the result of running Cython 3.0.8 locally to get things to be 3.12-compatible -- the rest is pretty straightforward I think.

prabhuramachandran (approved):
Wow, extensive changes. Thank you so much @larsoner, looks good to me.

prabhuramachandran:
> @prabhuramachandran when you get a chance can you
> 
> 1. (Review if you want and) approve this PR

Done.

> 2. Change the branch protection rules to reflect those in this PR

I don't see these options yet on the branch protection rules page, I suspect I will need to merge this first and do it.

> 3. Merge

Will do now.

PR #1301, opened 2024-05-23
Release 4.8.2

prabhuramachandran:


larsoner:
@prabhuramachandran did you happen to cythonize `array_ext.pyx` before release using Cython 3.0.10, and with NumPy 2.0 installed? I got a `diff` when I did. And without it, compiling against 2.0 fails. See the patch I had to add here:

https://github.com/conda-forge/mayavi-feedstock/pull/80

Might be worth cutting a 4.8.3 since the NumPy 2.0 release is imminent 😞

prabhuramachandran:
Oh no, I did not do that. Should I clobber the existing array_ext.c with this new version then? I can push a release in a few days I hope. Are there any other issues?

PR #1303, opened 2024-05-28
MAINT: Compile with Cython 3.0.10

larsoner:
Compiled while on NumPy 2.0rc2 (though I doubt this matters) and Cython 3.0.10.


larsoner:
@prabhuramachandran taking a stab at https://github.com/enthought/mayavi/pull/1301#issuecomment-2132411224, it seems like things are complicated if we try to support Python 3.8 since there is [no NumPy 2.0 for Python 3.8](https://pypi.org/project/numpy/2.0.0rc2/#files). Okay to bump to 3.9?

prabhuramachandran:
> @prabhuramachandran taking a stab at [#1301 (comment)](https://github.com/enthought/mayavi/pull/1301#issuecomment-2132411224), it seems like things are complicated if we try to support Python 3.8 since there is [no NumPy 2.0 for Python 3.8](https://pypi.org/project/numpy/2.0.0rc2/#files). Okay to bump to 3.9?

Thanks for doing this. Yes, I am perfectly fine with dropping 3.8 and moving to 3.9 for the tests. I just changed the branch protection rules but they don't seem to reflect on this PR. So I will just go ahead and merge this after a quick review.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-    "oldest-supported-numpy",
+    "numpy>=2.0.0rc2,<3",
```
prabhuramachandran, line 3: Is this a problem? Should we reduce this requirement, what if someone is using 1.21.6 wants to install mayavi?

prabhuramachandran (changes requested):
I am not sure about the change to the build requirements as this will force a new numpy even if someone does not have it.

pyproject.toml
larsoner: This is for the build system. So if someone does `pip install .` or whatever, it should do an isolated build using NumPy 2+. This should be fine for them since wheels compiled with NumPy 2 are backward compatible with NumPy 1.x.

prabhuramachandran: Yes, but since numpy 2.x does not work with 3.8, those on an older Python cannot use Mayavi. Do we want that or is Python 3.8 ancient history?

larsoner: Even though security support for 3.8 isn't over yet (it ends in 5 months), NumPy made the decision to drop it already. So to me it makes sense for Mayavi to drop it as well.

If it were trivial to support 3.8 I'd say "sure let's keep it", but I'm not sure how hard it would be to fix the errors seen here for example:

https://github.com/enthought/mayavi/actions/runs/9273915257/job/25515059891

```
  tvtk/src/array_ext.c:3467:13: error: call to undeclared function 'PyArray_MultiIter_ITERS'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
```

I tried for a little while but it wasn't trivial for me at least.

prabhuramachandran: Thanks for the clarification, in any case people could use the previous release of mayavi if they really want to use Python 3.8.

prabhuramachandran (approved):
Approving.

PR #1313, opened 2024-08-24
Added requirements files

vishalm:
1. Added requirements files
2. This is mac os based requirements files


prabhuramachandran:
Thank you for your PR. I am not sure we want to fix the dependencies like this. @larsoner do you have an opinion on this?

larsoner:
I have seen this used sometimes (e.g., in `pyvistaqt`), then dependabot comes through and updates something as a new release is made. Every once in a while something breaks. But in general in my experience it tends to create more busy work than help with progress. So I'm -0.5.

If we do add this, then we'd want to add a CI run that actually used the `requirements.txt`, and probably add a bot automerge (with suitable merge CI requirements) to mark any dependabot PR as automerge to reduce the maintenance burden. In other words, let dependabot open its PR, and it will get auto-merged if CIs are happy...

larsoner:
Closing since I don't think we'll go this route. Nowadays maybe `uv` could be used but I'm not sure it buys us too much

PR #1318, opened 2024-11-15
Recent versions of pyface don't have a pyface.__version__

mr-c:


prabhuramachandran (approved):
Thanks, looks good, could you please rebase this too.

PR #1329, opened 2024-12-17
MAINT: Require VTK 9+

larsoner:
Workaround for https://gitlab.kitware.com/vtk/vtk/-/issues/19562, but also pin fixes to a max version so that over time hopefully the shims die.

Speaking of which, it might be worth adding a min pin on VTK. Some code could be cleaned up by depending on vtk 9+, which is four years old now. Same thing with dropping Python 3.9, which most projects have done nowadays. Happy to add those here if desired.

*EDIT: Also restructures CIs explicitly to use latest VTK + NumPy + Qt bindings (preferring official PySide6) available, and comments about old ones.*

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      actions:
+        patterns:
+          - "*"
--- a/.github/workflows/run-mayavi-tests.yml
+++ b/.github/workflows/run-mayavi-tests.yml
@@ -62,7 +62,7 @@ jobs:
           - python-version: '3.10'
             qt-api: 'pyqt5'
             os: ubuntu-latest
-            vtk: 'vtk==9.2'
+            vtk: 'vtk==9.2.2'
           # In practice this one segfaults :(
           # - python-version: '3.9'
           #   qt-api: 'pyqt5'
--- a/.github/workflows/run-mayavi-tests.yml
+++ b/.github/workflows/run-mayavi-tests.yml
@@ -63,10 +63,11 @@ jobs:
             qt-api: 'pyqt5'
             os: ubuntu-latest
             vtk: 'vtk==9.2.2'
-          - python-version: '3.9'
-            qt-api: 'pyqt5'
-            os: ubuntu-latest
-            vtk: 'vtk==9.1.0'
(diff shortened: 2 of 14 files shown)


larsoner:
@prabhuramachandran okay to make the minimum VTK 9+? It is over 4 years old now, and the last 8.x release on PyPI only has wheels up through 3.7, which we don't even support anymore.

larsoner:
We would get even more cleanups if we required Python 3.10 and VTK 9.2 (2 years old) or at least VTK 9.1 (3 years old), but for now we at least get some by requiring VTK 9+.

larsoner:
Looks like in practice 9.1 segfaults:

https://github.com/enthought/mayavi/actions/runs/12400488425/job/34617800910?pr=1329

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      actions:
+        patterns:
+          - "*"
```
larsoner, line 10: Noticed this repo didn't have one of these but it's very convenient to have. It'll tend to run CIs once a month or every couple of months which is nice to have to check in. Not to mention actually having the actions stay up-to-date :)

larsoner:
Okay I think this one should be ready @prabhuramachandran !

prabhuramachandran:
> We would get even more cleanups if we required Python 3.10 and VTK 9.2 (2 years old) or at least VTK 9.1 (3 years old), but for now we at least get some by requiring VTK 9+ (4 years old).

I am not sure if there are any Enthought internal projects using older VTK. @dpinte, @mdickinson -- what do you think? Can we start requiring VTK 9+ and also Python 3.10? Thanks.

prabhuramachandran (approved):
Thanks, LGTM!

prabhuramachandran:
I think @dpinte is on leave but he did tell me that it is OK to move ahead. In any case for older versions, there are older releases of Mayavi.

prabhuramachandran:
@larsoner -- should I disable the required statuses before merging here?

prabhuramachandran:
@larsoner -- I will merge this for now, perhaps you can adjust the required tests if you have the permissions or let me know which ones you recommend that we keep.

larsoner:
@prabhuramachandran I don't have permissions to change the branch protection rules, but I would just set them to the set of runs that were in this PR when it was merged

PR #1342, opened 2025-04-18
Automatically exclude segfaulting vtk classes before setup

nauaneed:
- tvtk installation may break due to segfaulting vtk classes with vtk updates or on untested distros or python versions. This PR would be a mild relief to this issue.
- This PR makes the setup slower by a few seconds.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,6 +35,8 @@ exec(compile(open(fname).read(), fname, 'exec'), info)
 DEFAULT_HTML_TARGET_DIR = join('docs', 'build')
 DEFAULT_INPUT_DIR = join('docs', 'source',)
 
+from tvtk.filter_nosegfault import exclude_segfault_classes
+exclude_segfault_classes()
 
 class GenDocs(Command):
 
--- a/tvtk/_setup.py
+++ b/tvtk/_setup.py
@@ -8,7 +8,6 @@ import sys
 
 from setuptools import Distribution
 
-
 def can_compile_extensions():
     try:
         import numpy  # noqa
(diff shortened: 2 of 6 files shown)


nauaneed:
So, vtkXOpenGLRenderWindow gets excluded on ubuntu-latest, 3.12, pyside6, vtk>=9.4.

larsoner:
I like the idea in principle!

So far the (annoying) iteration has allowed us to identify and open several upstream bugs. So we'd want to make sure that running the setup has an easily parsable line saying which classes/methods segfault so we can open upstream bug reports if needed.

> So, vtkXOpenGLRenderWindow gets excluded on ubuntu-latest, 3.12, pyside6, vtk>=9.4.

It shouldn't, though, right? Assuming you have an X server running while generating at least for example it seems like it should work. And maybe it shouldn't be allowed to fail silently on most linux systems...

larsoner:
... okay it looks like `vtkXOpenGLRenderWindow` *should* probably be excluded. So maybe this is good enough to go once we make sure there is some mechanism / easy way to say which classes/methods etc. were excluded if possible. The current workflow I use is described here:

https://github.com/enthought/mayavi/blob/49e6c8ce9db68b7dbfc8d23dcee867e89c6f8ef7/tvtk/code_gen.py#L3-L15

Would be good to make sure that still works, or some variant still works to manually add excludes, open upstream bugs, etc.

nauaneed:
I was thinking purely as a user. One afternoon, I was annoyed at not being able to install this. My (_kind of narrow-minded_) thought process was that I, as a user, should be able to just get this installed at the cost of letting a few classes be excluded without caring about them. 

But I get the larger viewpoint from what you mentioned. It is necessary to know what exactly is broken and try to get it fixed upstream. 

How I am doing this is also very hacky. Feel free to close this PR :)

PR #1345, opened 2025-05-06
FIX: Fixes for VTK 9.4.2

larsoner:
Let's see if this works!

### DIFF
--- a/tvtk/code_gen.py
+++ b/tvtk/code_gen.py
@@ -4,7 +4,7 @@ This can be evoked for example by running from the ``mayavi`` root:
 
 ..code-block:: console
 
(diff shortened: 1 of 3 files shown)


larsoner:
Ugh seems like there is a bug with `vtkmodules.vtkRenderingOpenGL2.vtkXOpenGLRenderWindow` where it segfaults during the deconstructor and one with another class. Will require further investigation :(

prabhuramachandran (approved):
This looks good to me. We don't explicitly create the `XOpenGLRW` anywhere so this should not be an issue with the tests. The UI may run into problems, I will check this out after this PR is merged.

prabhuramachandran:
@larsoner -- so I tested this out and the UI is affected in that no traits specific to XOGLRW are exposed on the UI and only the generic OGLRW ones are which is perfectly OK to me. So I think we are good. Many thanks.

Issue #1348, opened 2025-05-29
`nan_color` Alpha Channel Incorrectly Applies Transparency to Non-NaN Data in Visualizations


sean-jost:
## Summary
When setting the `nan_color` attribute in Mayavi visualizations (e.g., `contour3d`, `surface`, `volume`), the alpha channel of `nan_color` incorrectly affects the transparency of active (non-NaN) cells, even when the dataset contains no NaN values. This occurs despite explicitly setting the `opacity` of the visualization to 1.0, leading to unintended transparency in the rendered output.

<img width="716" alt="Image" src="https://github.com/user-attachments/assets/1c4a5015-e0e7-4461-bfa1-0c53ae40e335" />

## Steps to Reproduce
1. Create a 3D dataset with no NaN values. For example:
   ```python
   import numpy as np
   from mayavi import mlab

   def create_sample_data(nx=25, ny=25, nz=15):
       data = np.zeros((nx, ny, nz))
       k_vals = np.linspace(0, 1, nz)
       data[:, :, :] = k_vals[None, None, :]
       return data

   data = create_sample_data(10, 10, 5)
   ```
2. Create two Mayavi figures to visualize the data using isosurfaces (or other methods like `surface` or `volume`):
   - In the first figure, set `nan_color` to `(0, 0, 0, 0)` (fully transparent).
   - In the second figure, set `nan_color` to `(0, 0, 0, 1)` (fully opaque).
   - Ensure the visualization’s `opacity` is set to 1.0.
   ```python
   # First figure (NaN color = transparent)
   mlab.figure(1, bgcolor=(1, 1, 1), size=(400, 600))
   vmin, vmax = np.nanmin(data), np.nanmax(data)
   contours = list(np.linspace(vmin, vmax, 4))
   src1 = mlab.contour3d(data, contours=contours, opacity=1.0, extent=[0, 25, 0, 25, 0, 15*2])
   src1.module_manager.scalar_lut_manager.lut.nan_color = 0, 0, 0, 0.9  # Nearly transparent
   src1.module_manager.scalar_lut_manager.lut_mode = 'blue-red'
   src1.module_manager.scalar_lut_manager.show_scalar_bar = True
   mlab.orientation_axes()

   # Second figure (NaN color = opaque)
   mlab.figure(2, bgcolor=(1, 1, 1), size=(400, 600))
   src2 = mlab.contour3d(data, contours=contours, opacity=1.0, extent=[0, 25, 0, 25, 0, 15*2])
[...]

PR #1353, opened 2025-07-08
Fix scene toolbar size and spacing.

drocheam:
The decorated_scene toolbar has unnecessary additional vertical spacing that arises from the display of empty tool names. You can see this spacing for instance in the images in https://github.com/enthought/mayavi/issues/1348 .

This is due to the default value of show_tool_names being set to True in:
https://github.com/enthought/pyface/blob/d6548833ee35e0d27b4db62316fbeef1a98497e5/pyface/action/i_tool_bar_manager.py#L32

This fix sets the value to False for the created ToolBarManager, therefore removing the spacing.


larsoner:
Makes sense, sorry for the long delay on this one!

PR #1364, opened 2025-10-10
Support VTK 9.5 and 9.6

larsoner:
Trying to add support. Now I see:
```
_________________________________________________________________ TestArraySource.test_2d_data _________________________________________________________________
mayavi/tests/test_array_source.py:102: in test_2d_data
    sc1 = surf.actor.mapper.input.point_data.scalars.to_array()
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E   AttributeError: 'NoneType' object has no attribute 'to_array'
```
So there is still some bug with defaults

Closes #1372
Closes #1371
Closes #1369
Closes #1367
Closes #1365
Closes #1363
Closes #1354
Closes #1352
Closes #1351
Closes #1350
Closes #1349
Closes #1346
Closes #1326
Closes #1324
Closes #1296

PR #1373, opened 2026-07-10
Drop C extension and move to pyproject.toml

larsoner:
Needs https://github.com/enthought/mayavi/pull/1364 to be merged then a rebase

1. Move to `pyproject.toml` + setuptools_scm for versioning (modernize project)
2. Get rid of NumPy C extension (speed gains not worth maintainability and installation issues)

I compared the contents before and after this switch and they match!


pyproject.toml
```diff
@@ -1,7 +1,105 @@
...
+    "importlib_resources; python_version<'3.11'",
+    "vtk",
+    "puremagic",
+]
+
+[project.optional-dependencies]
+app = ["envisage"]
+
+[project.urls]
+Homepage = "http://docs.enthought.com/mayavi/mayavi/"
```
prabhuramachandran, line 61: I think this is really not the homepage, perhaps github, or https://mayavi.readthedocs.io/en/latest/ could be used?

prabhuramachandran (commented):
Barring the minor homepage question/issue, it looks great otherwise. I think this will be very convenient. Thanks for the PR.

PR #1374, opened 2026-07-26
MAINT: Build wheel against latest VTK, test against older

larsoner:
Trying a strategy to distribute a `py3-<plat>` wheel: 

1. Build against newest VTK, one per OS (macOS, Linux, Windows)
2. Pin in the wheel `vtk<` that version (using dynamic dependencies in the wheel build)
3. Test in CIs that this wheel actually works for our test suite against the older VTK versions

This should make things easier on end users, assuming it works okay. I think we can rely on the CI scheme to tell us if it has a good chance of success. If it's a failure, we can revert. But I think it's worth trying!

### DIFF
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -1,0 +1,121 @@
+name: Wheel
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event_name }}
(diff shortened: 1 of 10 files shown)


larsoner:
Okay this seems to work -- building on 9.6.2 the library works on 9.2.5 on up! 🚀

PR #1375, opened 2026-07-27
Require VTK >= 9.4 (SPEC0), unify workarounds, and add CLAUDE.md

larsoner:
We had multiple places where we had "VTK does a bad thing let's fix it" code. This unifies them. Proof should be in the PRs. Claude Fable 5 drafted code changes and I reviewed.

Let's only support 2 years back (SPEC0) to keep things more maintainable, too -- allowing us to remove even more technical debt / code.


prabhuramachandran:
This looks good to me, and should be worth trying out! Thank you so much!

PR #1376, opened 2026-07-28
FIX: Fix bug with assertion

larsoner:
I (with Claude Opus 5) built a wheel with assertions on and verified the bug, then that this fixes the bug


larsoner:
Superseded by #1378

PR #1378, opened 2026-07-28
Test against prerelease wheel

larsoner:
- Add CI job against the latest prerelease wheel
- Add a scheduled run
- Open an issue if the run fails

Worked with Claude Opus 5 on this but I've done something similar in MNE-Python and other repos

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,6 +5,10 @@ concurrency:
 
 on:
   pull_request:
+  workflow_dispatch:
+  schedule:
+    # Weekly, mostly to catch breakage in VTK dev wheels
+    - cron: "38 4 * * 1"
 
 jobs:
   tests:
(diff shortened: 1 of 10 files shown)


opoplawski:
This appears to resolve the build issue, thanks!  Any chance of a new release soon?

larsoner:
Yes I'm hoping this week or next. One thing I'm considering is VTK 9.7 is on rc3, so waiting for its official release could be good. Is there a timeline/deadline on your (Fedora?) end?

opoplawski:
Mayavi has been in a fail to build state for a long time (https://koschei.fedoraproject.org/package/Mayavi) and is likely uninstallable in the latest release (F44).  At some point it will get dropped from the distribution though I think we still have some amount of time before that happens.

larsoner:
This is (at the end of the day) a pretty trivial fix so I'll go ahead and mark for merge-when-green!

PR #1380, opened 2026-07-30
Modernize and fix doc building

larsoner:
Deploys the docs to https://docs.enthought.com/mayavi/ on every push to `main`, and fixes enough of the docs pipeline to make that safe to run unattended.

Publishing has been a manual step at each release, so the live site currently shows 4.8.3. The new `docs.yml` builds both Sphinx trees, regenerates the mlab reference and the example gallery, and deploys via `actions/deploy-pages`.

### Fixes along the way

- **The build is warning-clean and stays that way.** Both Makefiles default `SPHINXOPTS` to `-nWT --keep-going`; the ~120 warnings that exposed are fixed at the source (duplicate labels from `include`d partials, a misspelled `:ref:` *label definition*, prose in bare `:type:` fields being cross-referenced, `wx`/`twisted` imports now mocked, and others).
- **The example gallery images are rendered on every run.** The datasets are cached, ImageMagick is replaced by Pillow, and HiDPI scaling is undone so a Retina Mac produces the same image sizes CI does. Cold run ~60 s, warm ~20 s.
- **Four examples that could no longer render are fixed:** `np.fromstring` binary mode (NumPy 2) in `canyon`/`canyon_decimation`, `scipy.special.sph_harm` (removed in SciPy 1.18) in `spherical_harmonics`, and the Python 2 `file()` builtin plus a data URL that now serves HTML in `chemistry`.
- **The mlab reference gets its illustrations back.** The generator derives image names from `module.__name__`, so it had been looking for `mayavi_mlab_*.jpg` while the committed files were `enthought_mayavi_mlab_*` — the package name that went away around 2010. Those pages have shipped with no figures for years: [mlab_helper_functions.html](https://docs.enthought.com/mayavi/mayavi/auto/mlab_helper_functions.html) documents `plot3d`, `points3d`, `surf`, `mesh` and the rest with not one picture between them, and builds with 13 after this change.
- **`mlab.test_simple_surf` rendered blank.** It draws the plane `z = x`, whose normal is exactly perpendicular to the default camera direction, so the surface was edge-on and invisible; its committed illustration is 1×1 px. `z = x + y` fixes it, and `trim()` now refuses to ship a figure that trimmed to nothing.
- **The changelog stops at 4.8.3.** `changes.html` now leads with a pointer to the [releases page](https://github.com/enthought/mayavi/releases) and keeps the older history below it, ahead of cutting 4.9.0.
- **Failures on `main` open an issue,** through a new `.github/actions/open-issue` composite action shared with the scheduled `tests.yml` run (whose inlined copy is replaced).
- **The site landing page** (`docs.enthought.com/mayavi/`, until now an unstyled `<h1>` and two links) links the docs' own stylesheets instead of restyling itself. That meant moving the Mayavi palette out of `_templates/layout.html` into `_static/mayavi.css`, which the docs load via `html_css_files` — the docs pages render pixel-identically.

### Notes for review

- Most of the diff is regenerated output, isolated in its own commit (`DOC: Update autogenerated docs`) — the hand-written changes are the other ~37 files.
- To read the rendered result, download the **`docs-site`** artifact from the run and serve it (`python -m http.server`). It is uploaded alongside the Pages artifact precisely because that one is a tar inside a zip. GitHub has no CircleCI-style linked preview — `deploy-pages` has a `preview` input, but it is alpha and not public.
- Eleven page URLs that were duplicate renderings of content inside `mlab.html` / `application.html` / `changes.html` are no longer built; each gets a meta-refresh + `rel=canonical` stub at its old address, aimed at the anchor the content now lives under.
- **Needs a settings change to take effect:** Pages is still set to deploy from the `gh-pages` branch, and has to be switched to "GitHub Actions" after this lands. Until then the deploy job will not publish. `gh-pages` becomes dead history and can be deleted.
- `pyproject.toml`'s `Documentation` URL moves from `mayavi.readthedocs.io` to `docs.enthought.com/mayavi`. The RTD project builds from `tkoyama010/mayavi-doc-translations`, a community mirror rather than anything this repo controls.

Generated with Claude Opus 5 under heavy supervision and iteration. At the end of the day the `-nWT --keep-going` means a lot is un-broken at the build end. And I'll download the artifacts and `python -m http.server` them to make sure they're okay 👍


larsoner:
Okay locally downloading the docs artifact, extracting, and running `python -m http.server` in the folder I get nice images:

### Better landing page

Screenshot 2026-07-29 at 10 21 15 PM

### Fixed images

Screenshot 2026-07-29 at 10 22 13 PM

docs/source/mayavi/auto/standalone.py
```diff
@@ -4,7 +4,7 @@
 
 """
 # Author: Prabhu Ramachandran <prabhu@aero.iitb.ac.in>
-# Copyright (c) 2007, Enthought, Inc.
+# Copyright (c) 2007-2020, Enthought, Inc.
```
prabhuramachandran: Should this be 2026? The copyright year is also something that ought to be either done away with or automated.

prabhuramachandran (commented):
This is wonderful, thank you so much!

docs/source/mayavi/auto/standalone.py
larsoner: Yeah this was autogenerated and carried over from another file. I'll open an issue about copyright unification so we don't forget, and at least fix this one instance

larsoner:
Switched! Let's see if it comes back green 🤞 

Screenshot 2026-07-30 at 2 20 07 PM

PR #1383, opened 2026-07-30
Fix messenger callback lifecycle handling

ShiroKSH:
## Summary

- Forward `obj_is_hash` from the public `disconnect` helper.
- Keep callbacks registered on live objects that evaluate false.
- Correct the parameter documentation for the `id`-keyed registry.

## Tests

- `python -m unittest tvtk.tests.test_messenger`


larsoner (approved):

larsoner:
Thanks @ShiroKSH !

PR #1384, opened 2026-07-30
Launch macOS Help without shell interpolation

ShiroKSH:
## Summary

- Launch macOS Help with an argument list instead of shell interpolation.
- Preserve paths and URLs containing spaces or shell metacharacters.
- Add a regression test for macOS launch arguments.

## Checks

- `python -m compileall -q mayavi/action/help.py mayavi/tests/test_action_help.py`
- Isolated `browser_open` regression check with the production module source.


mayavi/action/help.py
```diff
@@ -34,7 +34,7 @@
 
 def browser_open(url):
     if sys.platform == 'darwin':
-            os.system('open %s &' % url)
+        subprocess.Popen(['open', url])
```
larsoner: This without a `wait` leaves a zombie process. Maybe instead
```suggestion
        subprocess.run(['open', url])
```
?

mayavi/tests/test_action_help.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class TestBrowserOpen(unittest.TestCase):
+    def test_macos_passes_url_as_one_argument(self):
+        url = '/tmp/Mayavi Docs; unexpected-command'
+        with patch.object(action_help.sys, 'platform', 'darwin'):
+            with patch.object(action_help.subprocess, 'Popen') as popen:
+                action_help.browser_open(url)
+
+        popen.assert_called_once_with(['open', url])
```
larsoner:
```suggestion
            with patch.object(action_help.subprocess, 'run') as run:
                action_help.browser_open(url)

        run.assert_called_once_with(['open', url])
```

larsoner (changes requested):
Another option would be to drop the `darwin` conditional entirely here. I doubt it's needed anymore!

mayavi/action/help.py
ShiroKSH: Addressed in 0d44a79: `browser_open` now delegates to `webbrowser.open` on every platform, removing the custom subprocess path.

mayavi/tests/test_action_help.py
ShiroKSH: Addressed in 0d44a79: the regression test now mocks `webbrowser.open` and asserts the standard launcher call.

larsoner (approved):

PR #1385, opened 2026-07-31
Add multiple examples to docs

larsoner:
Add several more example images, plus fix a missing `r"`. Also cleans up `savefig` to use back-buffer capture, same as ParaView (made by VTK folks) and avoids raising the window.


docs/source/render_examples.py
```diff
@@ -39,6 +92,170 @@ def should_render(short_file_name, image_file):
...
+    def settle():
+        app = QtGui.QApplication.instance()
+        for _ in range(25):
+            app.processEvents()
+
+    mlab.show = lambda func=None: None
+    HasTraits.edit_traits = edit_traits
+    HasTraits.configure_traits = lambda self, *a, **k: edit_traits(self)
+    np.random.seed(0)
+    try:
```
prabhuramachandran, line 213: This is quite a brute-force approach. However, if the goal is to automatically get the screenshots perhaps this is necessary but all the monkey patching, exec, try blocks seem a bit heavy handed and makes me uncomfortable. Then again, the original also did a bit of this! I am just writing down my thoughts as I see the PR.

tvtk/pyface/tvtk_scene.py
```diff
@@ -796,11 +796,17 @@ def _create_control(self, parent):
...
+        # Read the back buffer, not the front one.  The front buffer is
+        # whatever the window server is showing, so anything overlapping the
+        # window ends up in the saved image -- which is why this used to raise
+        # the window first, stealing focus every time a figure was saved.  VTK's
+        # own regression tests and ParaView (both BSD-3-Clause) read the back
+        # buffer for the same reason:
+        # https://github.com/Kitware/VTK/blob/master/Testing/Rendering/vtkTesting.cxx
+        # https://gitlab.kitware.com/paraview/paraview/-/blob/master/Remoting/Views/vtkSMViewProxy.cxx
+        # _exporter_write holds the buffer swap off while the read happens.
+        w2if = tvtk.WindowToImageFilter(read_front_buffer=False)
```
prabhuramachandran, line 808: I tried this many many years ago but ran into some trouble, glad that it works well now.

tvtk/pyface/tvtk_scene.py
```diff
@@ -820,9 +826,14 @@ def _exporter_write(self, ex):
         else:
             aa_frames = rw.multi_samples
             rw.multi_samples = self.anti_aliasing_frames
+        # What is left in the back buffer after a swap is undefined by the
+        # OpenGL spec, so do not let one happen while the image is read.
+        swap_buffers = rw.swap_buffers
```
prabhuramachandran, line 825: I would suggest wrapping the change to `swap_buffers` in a try/finally. Otherwise if something went wrong in the render/update/write, the window will be in bad shape forcing a user to close the window and re-create it.

prabhuramachandran (dismissed):
This looks great overall, thanks for cleaning up and significantly improving  this long untouched piece of the infrastructure. The only comment that may need looking at is the try/finally for the swap_buffer attribute change. Thanks!

docs/source/render_examples.py
larsoner: Yeah this type of stuff I've done similarly in MNE for example to get 3D figures to "scrape" properly using sphinx-gallery and Pyvistaqt. Seems somewhat unavoidable to force Qt to actually show a window, settle geometry, paint, etc

tvtk/pyface/tvtk_scene.py
larsoner: Yeah I noticed the image quality goes up on macOS with this at least. Difference is subtle but I think it's a clear bugfix

larsoner:
Try/except implemented, and lots of other tiny little issues fixed. Once all examples run properly on CIs and the committed images have been pulled from the artifacts and committed (for consistency) I'll merge!

PR #1386, opened 2026-07-31
MAINT: Treat warnings as errors in the test suites

larsoner:
Allow VTK 9.7 by replacing the derived wheel cap (next minor of the build VTK) with an explicit MAX_VTK, and test NumPy nightlies alongside the VTK dev wheels so the same row covers both.

Then turn warnings into errors and deal with what that turns up:

- mayavi.core.common.exception() swallowed the failure and put up a modal pyface box, so a headed run blocked on a dialog and passed anyway; mayavi/tests/conftest.py re-raises instead.
- Contour and Volume assigned get_range's list to a Tuple trait, which warned and never compared equal, so the range update always fired.
- algs.min/max/mean/sum are deprecated as of VTK 9.6, which is also where numpy functions start dispatching on composite arrays.
- Introspecting every VTK getter (vtk_parser, tvtk_base.update_traits, tvtk_doc) inevitably calls deprecated ones, where the warning comes back out of the C wrapper as a SystemError.

The filters are per-suite, in each tests/conftest.py rather than pyproject.toml, so that they ship in the wheel and so also apply to the pytest --pyargs runs in wheel.yml.

integrationtests/ is not a pytest tree -- collecting it launches the Mayavi2 application and blocks on the window -- so it gets a conftest.py that hides it from a bare pytest.  Porting it is a follow-up.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -86,14 +86,19 @@ jobs:
         # --no-build-isolation, so it must be present in the environment
         python -m pip install --upgrade pip setuptools setuptools_scm wheel
         if [[ "${{ matrix.vtk }}" == "vtk-dev" ]]; then
-          # Dev wheels get their own pip call so that --pre only applies to VTK
+          # Dev wheels get their own pip calls so that --pre and the nightly
+          # indexes apply only to the package they belong to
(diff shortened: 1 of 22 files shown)


larsoner:
This one is pretty straightforward but @prabhuramachandran feel free to review if you want!

larsoner:
... happy to open a follow-up for any comments

prabhuramachandran:
LGTM.

PR #1387, opened 2026-07-31
MAINT: Treat warnings as errors when building the docs

larsoner:
-W only makes Sphinx's own diagnostics fatal, so nothing was watching the Python warnings from the half of the build that runs our code: the examples.  Make them errors in both places that run Python -- the two conf.py files for the Sphinx build, and scripts/render_docs.py for example execution, exported as PYTHONWARNINGS so the filters reach the per-example child processes as well.  capture_in_subprocess discards a child's output unless it exits non-zero, so a warning there was only ever going to be noticed by being raised, and RENDER_FAILURES then carries it to the end of the run.

What that turned up:

- volume_slicer_advanced compared a trait against a string literal with `is` (twice), which is a SyntaxWarning and was never true by luck.
- mlab_3D_to_2D called np.random.random_integers, long deprecated.
- the two julia_set examples let |z| overflow, which is the point of an escape-time iteration but should say so via np.errstate.
- the doc generators leaked a file handle per example -- several inside sort keys, so hundreds per build.  Read via pathlib, and build the generated pages in memory rather than into a file nobody closed.
- mlab_reference named its docutils components through the deprecated Publisher.set_components() rather than the constructor.
- the examples assigned to ndarray.shape, deprecated in NumPy 2.5.

The gallery reorders because it sorts by example length and the errstate wrapper made julia_set two lines longer.  No images are committed here: they render 3px taller off CI, so CI's own should be taken instead.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -94,6 +94,19 @@ The build is warning-clean and **kept** that way: both Makefiles default
 warning is genuinely unfixable, add it to `nitpick_ignore` in
 `docs/source/mayavi/conf.py` with a reason rather than dropping the flags.
 
+`-W` covers only Sphinx's *own* diagnostics, so **Python** warnings are made
+fatal separately, in the two places that run Python:
+
+- both `conf.py`s call `warnings.filterwarnings('error')`, for the Sphinx
+  build itself (autodoc importing our modules);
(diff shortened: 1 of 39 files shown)


prabhuramachandran:
LGTM.

PR #1389, opened 2026-07-31
DOC: Render the wx examples

larsoner:


docs/source/render_examples.py
```diff
@@ -273,9 +282,112 @@ def settle():
...
+    np.random.seed(0)
+    try:
+        exec(compile(Path(filename).read_text(), filename, 'exec'),
+             {'__name__': '__main__', '__file__': os.path.abspath(filename)})
+        if not frames:
+            raise RuntimeError('%s opened no frame' % filename)
+        frame = frames[-1]
+        frame.Show()
+        app = wx.GetApp()
+        for _ in range(30):
```
prabhuramachandran: Why the magic number 30?

prabhuramachandran (commented):
LGTM barring the magic for loop running 30 times on Yield?

docs/source/render_examples.py
larsoner: Cruft, I'll push a commit that waits on a condition with a deadline instead! Has to do with X being asynchronous so you have to wait (im)patiently for it to actually create windows. But explicitly checking in a loop should be better than picking an arbitrarily high number

PR #1393, opened 2026-07-31
FIX: Let font_family take VTK_FONT_FILE

### LINKED ISSUE #1391: font_family trait rejects VTK_FONT_FILE (4) on VTK ≥ 9.4 — breaks update_traits() resync, not just direct assignment
## Summary

`font_family` trait's valid range no longer includes `VTK_FONT_FILE` (4) on VTK ≥ 9.4, breaking both direct assignment and the `update_traits()` resync path for anyone using a custom TrueType font.

## Details

On VTK 9.3, `tvtk.TextProperty().font_family = 4` (`VTK_FONT_FILE`) worked to select a custom font via `font_file`. On VTK 9.6, the trait's valid range no longer includes `4`:

```python
>>> tp.font_family = 4
TraitError: The 'font_family' trait of a TextProperty instance must be
'arial' or 'courier' or 'times' (or any unique prefix) or 0 or 1 or 2,
but a value of 4 <class 'int'> was specified.
```

Setting the underlying VTK object directly (bypassing the trait) works and renders correctly:

```python
prop._vtk_obj.SetFontFamily(4)
prop._vtk_obj.SetFontFile("/path/to/font.ttf")
```

**However**, this only avoids the *initial* error. Any subsequent change to **any other trait** on the same `TextProperty` (e.g. `prop.bold = False`) triggers `_do_change()` → `update_traits()`, which re-reads every property from the real VTK object and calls `setattr()` back onto the traits — including `font_family`, now genuinely `4` on the C++ side. This hits the same broken validation range and raises `TraitError` again inside a traits notification handler, on *every* such change (non-fatal — swallowed/printed — but noisy, and clearly a bug):

```
Exception occurred in traits notification handler for object: vtkTextProperty (...)
  ...
  FontFamilyAsString: File
  FontFile: /usr/share/fonts/truetype/liberation/LiberationMono-Regular.ttf
  ...
traits.trait_errors.TraitError: The 'font_family' trait of a TextProperty
instance must be 'arial' or 'courier' or 'times' (or any unique prefix)
or 0 or 1 or 2, but a value of 4 <class 'int'> was specified.
```

## Likely root cause

VTK moved `FontFamily`'s real storage from a plain clamped `int`
(`vtkSetClampMacro(FontFamily, int, VTK_ARIAL, VTK_TIMES)`) to a string
(`FontFamilyAsString`, via `vtkSetStringMacro`/`vtkGetStringMacro`), keeping
`SetFontFamily(int)` / `GetFontFamily()` only as thin compatibility wrappers.

Whatever tvtk's generator used to build the `font_family` trait's valid-value
range for VTK 9.3 apparently sourced it from the old macro form, and no longer
detects `VTK_FONT_FILE` now that the macro is gone.

## Reproduction

```python
from tvtk.api import tvtk

tp = tvtk.TextProperty()
tp._vtk_obj.SetFontFamily(4)
tp._vtk_obj.SetFontFile("/path/to/font.ttf")
tp.bold = False   # triggers update_traits() -> TraitError on font_family resync
```

## Environment

- VTK: 9.6.2
- tvtk / mayavi: built from `main` (post VTK 9.6 / Python 3.14 support work)
- Python: 3.13

---

Happy to help narrow down where in the generation pipeline the valid-value range gets built, if that's useful.

larsoner:
`_gen_state_methods()` builds a state trait's set of valid values from the `SetXToY` methods alone, so any value `SetX()` accepts without a matching `SetXToY` is missing from the map.  `vtkTextProperty` is the live case: it has no `SetFontFamilyToFile()`, yet `VTK_FONT_FILE` (4) is how a custom `FontFile` is selected, and out-of-range values clamp to `VTK_UNKNOWN_FONT` (3).  Both were rejected by the trait, so `update_traits()` raised inside a traits notification handler on every resync of a text property that really was set to a font file -- and setting it through tvtk at all was impossible.

Add the two names to the map, as `vtkPoints.DataType` already does for `VTK_ID_TYPE` on Windows.

Closes #1391

### DIFF
--- a/tvtk/tests/test_tvtk.py
+++ b/tvtk/tests/test_tvtk.py
@@ -195,6 +195,14 @@ class TestTVTK(unittest.TestCase):
         obj.SetSpecularColor(val)
         self.assertEqual(p.specular_color, val)
 
+        # A state value with no SetXToY method to discover it from must
+        # still resync, else this raises (see tvtk/WORKAROUNDS.md).  #1391
+        tp = tvtk.TextProperty()
+        tp._vtk_obj.SetFontFamily(4)  # vtk.VTK_FONT_FILE
+        self.assertEqual(tp.font_family, 'file')
+        tp.font_family = 0  # vtk.VTK_ARIAL
+        self.assertEqual(tp._vtk_obj.GetFontFamily(), vtk.VTK_ARIAL)
+
     def test_obj_del(self):
         """Test object deletion and reference cycles."""
         p = tvtk.Property()
--- a/tvtk/tests/test_tvtk.py
+++ b/tvtk/tests/test_tvtk.py
@@ -182,7 +182,7 @@ class TestTVTK(unittest.TestCase):
     def test_auto_update(self):
         """Test if traits are updated when the VTK object changes."""
         p = tvtk.Property()
-        obj = p._vtk_obj
+        obj = tvtk.to_vtk(p)
         obj.SetEdgeVisibility(1)
         self.assertEqual(p.edge_visibility, 1)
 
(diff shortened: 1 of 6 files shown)


tvtk/tests/test_tvtk.py
```diff
@@ -195,6 +195,14 @@ def test_auto_update(self):
         obj.SetSpecularColor(val)
         self.assertEqual(p.specular_color, val)
 
+        # A state value with no SetXToY method to discover it from must
+        # still resync, else this raises (see tvtk/WORKAROUNDS.md).  #1391
+        tp = tvtk.TextProperty()
+        tp._vtk_obj.SetFontFamily(4)  # vtk.VTK_FONT_FILE
```
prabhuramachandran: Ideally one should get the underlying VTK object via `tvtk.to_vtk` and reverse with `tvtk.to_tvtk`. Accessing the private `_vtk_obj` is not clean.

prabhuramachandran (changes requested):
Just a minor change.

tvtk/tests/test_tvtk.py
larsoner: Okay, pushed a commit to fix usages of `_to_vtk` here and more generally (since things weren't very consistent -- some even showed up in examples!)

prabhuramachandran (approved):
Thank you!

Issue #1397, opened 2026-08-01
Wayland has to use X fallback for rendering


larsoner:
We currently have to do QT_QPA_PLATFORM=xcb (or users need to do it) to get things to work there. We'll need to adopt something like https://github.com/pyvista/pyvistaqt/pull/848 or (probably) https://gitlab.kitware.com/vtk/vtk/-/merge_requests/9443 to get Wayland support working. Ideally the VTK folks would get this working and we would purely vendor, but as Wayland popularity grows, we might need to try one of these ourselves directly instead 😅

See for example #1396

PR #1398, opened 2026-08-01
FIX: Improve Wayland support

### LINKED ISSUE #1396: Fresh install of mayavi `main` (post VTK 9.6 / Python 3.14 support work) crashes
## Summary

Fresh install of mayavi `main` (post VTK 9.6 / Python 3.14 support work) crashes on **any** attempt to create a Mayavi scene widget (`SceneEditor` + `MlabSceneModel`), with an X11 `BadWindow` error — reproducible with a minimal script, independent of GPU driver or rendering backend.

## Details

```python
from traits.api import HasTraits, Instance
from traitsui.api import View, Item
from mayavi.tools.mlab_scene_model import MlabSceneModel
from mayavi.core.ui.mayavi_scene import MayaviScene
from tvtk.pyface.scene_editor import SceneEditor

class MinimalTest(HasTraits):
    scene = Instance(MlabSceneModel, ())
    view = View(Item('scene', editor=SceneEditor(scene_class=MayaviScene),
                      show_label=False, resizable=True),
                resizable=True)

MinimalTest().configure_traits()
```

crashes at the point the scene widget is shown:

```
QSocketNotifier: Can only be used with threads started with QThread
X Error of failed request:  BadWindow (invalid Window parameter)
  Major opcode of failed request:  3 (X_GetWindowAttributes)
  Resource id in failed request:  0x3
  Serial number of failed request:  7
  Current serial number in output stream:  8
```

The standalone `mayavi2` app crashes the same way (on a real, physical local X session, no remote/forwarded display), both before and after patching an unrelated `envisage.plugins.python_shell` bug encountered along the way.

## What's been ruled out

- **Not remote/forwarded display**: reproduced on a physical machine, X running on real hardware, direct local session.
- **Not GLX indirect rendering**: `direct rendering: Yes` confirmed via `glxinfo`.
- **Not a GPU/driver capability issue**: reproduces identically whether using the `nouveau` driver (hardware, OpenGL 4.3 Compatibility Profile) or forced software rendering via `LIBGL_ALWAYS_SOFTWARE=1` (Mesa llvmpipe). Same crash, same resource ID, same serial numbers, in both cases.
- **Not `envisage`/workbench-specific**: reproduces with the minimal script above, which uses plain TraitsUI + `SceneEditor`, no `envisage.ui.workbench` involved at all.

## Suspected root cause

This may be related to the change discussed in #1345, where the comment thread notes:

> "We don't explicitly create the XOpenGLRW anywhere so this should not be an issue with the tests. The UI may run into problems, I will check this out after this PR is merged."
>
> "@larsoner -- so I tested this out and the UI is affected in that no traits specific to XOGLRW are exposed on the UI and only the generic OGLRW ones are which is perfectly OK to me."

If VTK's automatic render-window-type selection (rather than an explicit `vtkXOpenGLRenderWindow`) now picks a different native window creation path under this environment (Python 3.14 / Qt6 / VTK 9.6), that could explain why Qt's window embedding (`QVTKRenderWindowInteractor` / `SceneEditor`) ends up querying attributes on a window ID that was never created as a conventional X11 window — consistent with the crash happening on the very first render-window-related X request (low resource ID, low serial number), regardless of which GL driver/backend is behind it.

## Environment

- VTK: 9.6.2
- mayavi / tvtk: built from `main` (post VTK 9.6 / Python 3.14 support work)
- Python: 3.14
- OS: Debian (unstable)
- Qt binding: PyQt5
- GPU: NVIDIA (Maxwell, via `nouveau`) — same crash reproduced with `LIBGL_ALWAYS_SOFTWARE=1` (llvmpipe)

## Question

Is there a way to force the classic `vtkXOpenGLRenderWindow` creation path explicitly, to check whether that restores the previous (working) behavior? Happy to test on my end if there's a suggested override.

larsoner:
Set QT_QPA_PLATFORM=xcb from `mayavi/__init__.py` and `tvtk/pyface/__init__.py`, before pyface creates the QApplication as an import side effect of `traitsui.api`. raise `RuntimeError` from the scene widget when the app predates `mayavi` and is already on wayland. Proper support tracked in #1397

Closes #1396


tvtk/qt_x11.py
```diff
@@ -0,0 +1,69 @@
...
+import os
+import sys
+
+
+def x11_platform_override(platform, environ):
+    """The Qt platform to request so VTK render windows can embed, or None.
+
+    Pure logic split out for testing; see the module docstring for why.
+    """
+    if platform != 'linux':
```
prabhuramachandran: What if this were on a bsd system? Instead why not check for "not Windows/MacOS"?

tvtk/qt_x11.py
```diff
@@ -0,0 +1,69 @@
...
+        return None
+    if environ.get('ETS_TOOLKIT', 'qt4') not in ('qt4', 'qt', ''):
+        return None  # wx or null toolkit: leave Qt alone
+    if 'QT_QPA_PLATFORM' in environ:  # an explicit choice wins
+        return None
+    if 'WAYLAND_DISPLAY' not in environ:  # not a Wayland session
```
prabhuramachandran, line 41: Given that this is checked, I am not sure if the platform check of linux is necessary at all?

tvtk/qt_x11.py
```diff
@@ -0,0 +1,69 @@
...
+    ``platform_name`` is ``QGuiApplication.platformName()`` of the running
+    application, so this can only be checked -- and the clear error raised in
+    place of the Xlib abort -- once steering is already too late.
+    """
+    if platform == 'linux' and platform_name.startswith('wayland'):
+        return (
+            "A VTK render window cannot be embedded in a Qt window on the "
+            "'%s' platform: VTK draws to a native X11 window, and the window "
+            "ids Qt fabricates on Wayland have no X11 window behind them, so "
+            "VTK's first X request on one fails with BadWindow.  Set the "
```
prabhuramachandran: Is the only case that this happens when the user explicitly set QT_QPA_PLATFORM to something apart from xcb?

I am not sure if 99% of the audience will get the details, instead we can simply say "We detect that you are on Wayland and have please set the environment variable  QT_QPA_PLATFORM=xcb and try again."

tvtk/qt_x11.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def embedding_error(platform, platform_name):
+    """Why a VTK render window cannot embed in a Qt window here, or None.
+
+    ``platform_name`` is ``QGuiApplication.platformName()`` of the running
+    application, so this can only be checked -- and the clear error raised in
+    place of the Xlib abort -- once steering is already too late.
+    """
+    if platform == 'linux' and platform_name.startswith('wayland'):
```
prabhuramachandran: Again, should we worry about bsd here?

prabhuramachandran (changes requested):
Thanks for this, I've made some comments that may require some care. I think the error message can be simplified for the user.

tvtk/qt_x11.py
larsoner: The other case is a `QApplication` that already exists when `mayavi` is imported, locking the platform before this function could implement the workaround. That's why the check exists during widget-construction time as well (otherwise people can get cryptic errors that are hard to debug like the original report)

larsoner:
Simplified/fixed: handling across OSes and error message

prabhuramachandran (approved):

PR #1401, opened 2026-08-01
FIX: Wrap the vtkDataSetAttributes Copy* flags as toggle traits (#1355)

### LINKED ISSUE #1355: Can't set cell_data scalars in UnstructuredGrid
Hello,

I have some code that I use to write an UnstructuredGrid mesh to vtk. I have been using Python 3.10, mayavi 4.8.2, and vtk 9.3.0 for some time without issues. I recently upgraded the mayavi version to 4.8.3 and vtk to 9.5.0 and get the following error  after the code below         

```
        unstructured_grid = tvtk.UnstructuredGrid(points=points)
        unstructured_grid.set_cells(cell_types, offset, cell_array)
       
        unstructured_grid.cell_data.sclars = cell_data  # <-- error occurs at this line when trying to access UnstructuredGrid.cell_data
        unstructured_grid.cell_data.sclars.name = 'MaterialID'

        writer = tvtk.UnstructuredGridWriter(file_name=self.filename)
        writer.set_input_data(unstructured_grid)
        writer.update()
        writer.write()
```

`traits.trait_errors.TraitError: Cannot set the undefined 'copy_global_ids' attribute of a 'CellData' object.`

Any help would be awesome! I tried installing mayavi from the master brance (I think the version is 4.8.4dev0) and the issue sticks around. I also get the same error if I try python 3.12

larsoner:
- a pre-2011 blocklist in `VTKMethodParser._find_toggle_methods` kept `CopyScalars/Vectors/Normals/Tensors/TCoords/GlobalIds/PedigreeIds` from wrapping as On/Off toggles ("wrap badly", reason lost to history);
- the VTK 9.5 support work (#1364) then also skipped their Get/Set pair in _find_get_set_methods to silence the resulting update_traits() crash, which deleted the traits outright.

This `blocklist` is obsolete.  Remove both skips so all eleven Copy* flags wrap properly. Tested on 9.6.2, on VTK 9.4.x with the 9.6.2 wheel, and on 9.4.x with a 9.4.x-generated wheel.

Closes #1355

### DIFF
--- a/tvtk/WORKAROUNDS.md
+++ b/tvtk/WORKAROUNDS.md
@@ -58,9 +58,8 @@ from crashing or producing garbage:
   property getters).
 - `_find_get_set_methods()`: per-class/method skips for getters that hang or
   wrap badly (`vtkDataEncoder`/`vtkWebApplication`,
-  `vtkPiecewisePointHandleItem`, the `vtkDataSetAttributes.Copy*` pair on
-  VTK 9.5.2), and version-keyed "broken getter -> no default" entries near
-  the end (grep `Broken in`).
+  `vtkPiecewisePointHandleItem`), and version-keyed "broken getter -> no
+  default" entries near the end (grep `Broken in`).
 - `BROKEN_GETTERS` (module level): getters whose VTK-documented precondition
   a fresh object violates, so probing them aborts an assertion-enabled build.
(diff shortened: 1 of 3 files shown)


prabhuramachandran:
LGTM, thanks, merging.

PR #1402, opened 2026-08-01
Fix 1276 qt5 wheel zoom

### LINKED ISSUE #1276: QVTKRenderWindowInteractor.py _emit_wheel_event() will produce a name error QWheelEvent undefined.
The import probably got accidentally dropped when Qt6 logic was added. This only happens if pyface.is_qt4 is True which nowadays is quite unlikely.

larsoner:
Closes #1276 

Drop PyQt4 and PySide support, continue supporting PyQt5 and PySide2 (maybe last release for those two?). Pull in upstream fixes from VTK gitlab. Fix the aforementioned issue.


prabhuramachandran:
LGTM, thanks, merging.

PR #1405, opened 2026-08-02
FIX: Make savefig honor the requested size on Qt

### LINKED ISSUE #1288: mlab.savefig saves wrong resolution (for first two images in loop)
When using `mlab.savefig(<imagename>, size=(x, y))` the resolution of the saved image is about 10x that specified: 
```python
from mayavi import mlab

f = mlab.figure()
mlab.test_plot3d()
mlab.savefig('savetest.png', size=(100,80))
```
resolution:
```
$ identify -format '%f %w x %h \n' savetest.png 
savetest.png 1200 x 891 
```

When using `mlab.savefig()` in a for loop to generate a series of images, the first two will have this wrong resolution, the rest will be correct. 
MWE: 
```python
from mayavi import mlab

f = mlab.figure()
for n in range(10):
    mlab.test_contour3d()
    mlab.draw()
    mlab.savefig(f'iteration_{n}.png', size=(100, 80))
    mlab.clf()
```

The resolutions are (assuming ImageMagick is installed)

```bash
$ identify -format '%f %w x %h \n' iteration_*
iteration_0.png 1200 x 891 
iteration_1.png 1200 x 891 
iteration_2.png 99 x 78 
iteration_3.png 99 x 78 
iteration_4.png 99 x 78 
iteration_5.png 99 x 78 
iteration_6.png 99 x 78 
iteration_7.png 99 x 78 
iteration_8.png 99 x 78 
iteration_9.png 99 x 78 
```

This also occurs when using the virtual frame buffer [http://docs.enthought.com/mayavi/mayavi/tips.html#rendering-using-the-virtual-framebuffer](url). 

when using the direct frame buffer, I see the rendering process uses the whole window, and switches to a much smaller section after the third image.

larsoner:
Closes #1288

A layout-settling approach adapted from mne-tools/mne-python#14087 (BSD 3-clause; known to work there in tests and doc builds). The wx backend has the same latent `set_size` problem, deliberately left alone because it would be a pain to dig into (and Qt I expect is much more widely used).

### DIFF
--- a/mayavi/tests/test_savefig_size.py
+++ b/mayavi/tests/test_savefig_size.py
@@ -1,0 +1,71 @@
+"""Regression test for mlab.savefig(size=...) on an onscreen figure (gh-1288).
+
+The first savefig used to capture at the window's size instead of the
+requested one: set_size resized the embedded scene widget, which the
+enclosing layout immediately reasserted, and the pre-layout widget size
+inflated the auto magnification.
+"""
+# Copyright (c) Enthought, Inc.
+# License: BSD Style.
+
+import os
+import shutil
+import subprocess
+import sys
+
+import pytest
+
+_SCRIPT = """
+import os
+import tempfile
+
+from mayavi import mlab
+from tvtk.api import tvtk
+
+
+def png_size(fname):
+    r = tvtk.PNGReader(file_name=fname)
+    r.update()
+    ext = r.output.extent
+    return (ext[1] + 1, ext[3] + 1)
+
+
+d = tempfile.mkdtemp()
+mlab.figure()
+sizes = []
+for n in range(3):
+    mlab.test_contour3d()
+    mlab.draw()
+    fname = os.path.join(d, 'i%d.png' % n)
+    mlab.savefig(fname, size=(100, 80))
+    mlab.clf()
+    sizes.append(png_size(fname))
+# the first save is the regression: it saw the unsettled window size
+assert sizes == [(100, 80)] * 3, sizes
+
+# larger than the window exercises the magnification path
+big = os.path.join(d, 'big.png')
+mlab.savefig(big, size=(1200, 900))
+assert png_size(big) == (1200, 900), png_size(big)
+print('OK')
+"""
+
+
+@pytest.mark.timeout(180)
+def test_savefig_size_onscreen():
+    if sys.platform != 'linux':
+        pytest.skip('needs xvfb-run or an X display; exercised on Linux')
+    from traits.etsconfig.api import ETSConfig
+    if ETSConfig.toolkit not in ('', 'qt', 'qt4'):
+        pytest.skip(f'Qt-only test, got toolkit={ETSConfig.toolkit!r}')
+    cmd = [sys.executable, '-c', _SCRIPT]
+    xvfb = shutil.which('xvfb-run')
+    if xvfb is not None:
+        # a fresh server keeps windows off the developer's desktop, and
+        # nests fine inside CI's own Xvfb
+        cmd = [xvfb, '-a'] + cmd
+    elif not os.environ.get('DISPLAY'):
+        pytest.skip('no display and no xvfb-run')
+    res = subprocess.run(cmd, capture_output=True, text=True, timeout=150)
+    assert res.returncode == 0, res.stdout + res.stderr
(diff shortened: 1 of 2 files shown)


mayavi/tests/test_savefig_size.py
```diff
@@ -0,0 +1,71 @@
...
+
+def png_size(fname):
+    r = tvtk.PNGReader(file_name=fname)
+    r.update()
+    ext = r.output.extent
+    return (ext[1] + 1, ext[3] + 1)
+
+
+d = tempfile.mkdtemp()
+mlab.figure()
```
prabhuramachandran: Should we add an explicit size here instead of relying on the default which is 400x350?

tvtk/pyface/ui/qt4/scene.py
```diff
@@ -358,15 +358,76 @@ def render(self):
...
+
+        Qt only settles geometry when the posted LayoutRequest events are
+        delivered, and every layout in the chain has to be activated, not
+        just the window's.  Activating them directly is synchronous and,
+        unlike pumping the event loop, cannot run arbitrary queued events
+        from inside a save.  Adapted from mne-tools/mne-python#14087.
+        """
+        widget = self._vtk_control
+        processed = set()
+        while widget is not None and widget not in processed:
```
prabhuramachandran, line 379: Should the widget not in processed be in the `if layout is not None`? If the widget is processed, then this will exit the while, is that what you want?

tvtk/pyface/ui/qt4/scene.py
```diff
@@ -358,15 +358,76 @@ def render(self):
...
+
+    def _realize(self):
+        """Flush a pending show so the widget geometry is meaningful.
+
+        Until the queued show/layout events have run once, the widget
+        chain reports pre-layout default sizes (and activating the layouts
+        just bakes them in).  This pumps the event loop only in that
+        never-shown state, so steady-state calls stay purely synchronous.
+        """
+        for _ in range(10):
```
prabhuramachandran: Don't like this magic 10 number.

tvtk/pyface/ui/qt4/scene.py
```diff
@@ -358,15 +358,76 @@ def render(self):
...
+            # framebuffer at the old size (enthought/mayavi#1288).  Resize
+            # the top-level window and nudge until the layout lands this
+            # widget at the requested size -- the first pass absorbs the
+            # decoration/toolbar delta, so this converges in one or two.
+            self._realize()
+            self._activate_layouts()
+            for _ in range(3):
+                err_w = size[0] - ctrl.width()
+                err_h = size[1] - ctrl.height()
+                if not (err_w or err_h):
```
prabhuramachandran: Could this be instead `if not ((err_w > 0) or (err_h > 0))`? Just to be explicit since they are ints and not booleans?

prabhuramachandran (commented):
A few minor points for your consideration.

tvtk/pyface/ui/qt4/scene.py
larsoner: It should exit if you hit one that is already in the `processed` set... it's a bit of extra paranoia because you shouldn't have this sort of circular parenting in principle I think. But this should work as-is

tvtk/pyface/ui/qt4/scene.py
larsoner: In MNE we went with 3... in practice it was almost always (but not always!) 1. I don't like magic numbers either but I'm not sure if there's a good alternative!

larsoner: (... I like the alternative of a bare `while <not the right size>:` much worse... infinite loops are bad news)

tvtk/pyface/ui/qt4/scene.py
larsoner: They could be greater or less in practice so it would need to be `!= 0`. But I can make that part explicit

tvtk/pyface/ui/qt4/scene.py
larsoner: Actually this constant is different... but I'll name it and add a comment along the way!

prabhuramachandran:
LGTM, thanks! Merging.

PR #1406, opened 2026-08-02
FIX: Read the front buffer actually

larsoner:
Reading the back buffer instead throws away anti-aliasing that `_exporter_write` bumps `anti_aliasing_frames` up to get.

This took some painful digging... `vtkOpenGLRenderWindow::ReadPixels` takes two different paths.  A front read returns the `DisplayFramebuffer`, which `Frame()` has already resolved with a gamma-correct shader. A back read returns the raw multisample framebuffer, which it resolves inline with a plain `glBlitFramebuffer` (naive) average.  The two disagree on every anti-aliased edge pixel.

With VTK 9 a front read returns VTK's own framebuffer, not whatever the window server is showing, so nothing overlapping the window can end up in the saved image and `_lift()` stays the no-op it now is.

Found while banging my head against an equivalent change in PyVista https://github.com/pyvista/pyvista/pull/8858. I'm going to look into the VTK end and see if it's something that can/should be unified.


mayavi/sources/vrml_importer.py
```diff
@@ -116,7 +116,7 @@ def _update_reader(self):
                or len(reader.file_name) == 0:
             return
         actors1 = [x for x in self.scene.renderer.actors]
-        reader.read()
+        reader.update()
```
larsoner, line 119: I pushed some fixes to ensure exceptions were reraised in all tests (including integration tests) and it found a bug in this importer, so it tags along for the ride on this PR

prabhuramachandran:
LGTM, merging.

PR #1407, opened 2026-08-03
Add codecov coverage

larsoner:
Let's see what our numbers are actually like!


pyproject.toml
```diff
@@ -109,6 +110,63 @@ namespaces = true
...
+[tool.coverage.report]
+ignore_errors = true
+# `exclude_also` adds to coverage's defaults; the `exclude_lines` this grew out
+# of replaced them, which is why "pragma: no cover" had to be restated there
+exclude_also = [
+    "def __repr__",
+    "raise AssertionError",
+    "raise NotImplementedError",
+    "if __name__ == .__main__.:",
+    "if TYPE_CHECKING:",
```
prabhuramachandran, line 166: Just curious where these come from? We don't have any `TYPE_CHECKING` or use of `@overload`?

prabhuramachandran:
LGTM, merging for now. Just one minor comment which can be addressed later.

PR #1408, opened 2026-08-04
Don't set the default size of the scene.

prabhuramachandran:
Doing this messes up the relative sizes of the dialog. Instead setting an ID will ensure that any user defined size will be reused on the next launch.


larsoner:
With these doc commits I've been pulling the CI-generated images after a run and committing them. Has the bonus side effect of showing the improvement here:

image

larsoner:
... marking for merge-when-green, thanks in advance @prabhuramachandran !

larsoner:
... and I see now that a number of images accidentally got pushed at a large size, which has now been fixed.

PR #1412, opened 2026-08-05
Improve coverage

larsoner:
Improve our coverage by tracking code gen, and a few big targets. Caught a couple minor bugs along the way!

Also remove `runtests.py` and its related entry points -- nowadays people should invoke `pytest` directly for anything test-related. This is the proper way to ensure fixtures etc. are available. It still looked for `nose`...


larsoner:
@prabhuramachandran this also now fixes a minor bug with using 9.6.2 wheels on VTK 9.4. Maybe in a week or two once the dust settles we should cut a quick 4.9.1 (should be easy enough now that it's clicking on GH!).

larsoner:
Okay this keeps things green and is mostly a coverage improvement so I'll go ahead and merge

PR #1414, opened 2026-08-06
FIX: Make "View -> Other... -> Python" work again

larsoner:
@prabhuramachandran this should hopefully fix your bug. This is a workaround for envisage, I should probably open an upstream issue about this


prabhuramachandran:
I tested and the issue is fixed for me but I am not clear if the builds are failing because of this. Without this fix running the advanced_visualization examples fail. I am not sure how the docs ran before this fix?

larsoner:
Okay works on both envisage 8.0.0 and on https://github.com/enthought/envisage/pull/639 which fixes the problem upstream

prabhuramachandran:
@larsoner -- Given that this is now fixed in upstream and released as well, can we close this PR?

larsoner:
It'll still be broken for anyone on 8.0.0 or earlier, but that's probably not so many (and the fix is "update envisage"), so okay with me to close!

